PR #3, opened 2015-03-24
Try to use http2

justinfagnani:
This does not work yet...


justinfagnani:
Added new non-expired self-signed certs, but it's still not working on Chrome Canary. Failing with ERR_SPDY_INADEQUATE_TRANSPORT_SECURITY

PR #6, opened 2015-04-03
Make port configurable, plus some cleanup

cdata:
- The port can now be configured with `-p` when using CLI
 - A bug was fixed that caused the current component directory to be console
   logged as `undefined`.
 - Some code was factored into different modules for clarity.


justinfagnani:
:surfer:

PR #9, opened 2015-04-16
change the start message to be a valid URL

garlicnation:
This makes it easier for shells or logreaders to pick up the URL.


antono:
:+1:

PR #11, opened 2015-04-21
Redirect root to target component directory

mbleigh:
I didn't actually run this locally, but it ought to work :)

### DIFF
--- a/src/make_app.js
+++ b/src/make_app.js
@@ -31,6 +31,10 @@ function makeApp(componentDir, packageName, headers) {
   console.log('Serving components from ' + componentDir);
 
   var app = express();
+  app.get('/', function (req, res) {
+    // redirect homepage to 
+    res.redirect(301, '/components/' + packageName);
+  });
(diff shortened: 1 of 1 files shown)


justinfagnani:
Sorry for the delay, I somehow didn't have notifications on for this repo.

I'll try this locally, but I also have another change sitting around which ads a landing page with links into files and eventually documentation. That might be a good place to hang any new features off of. WDYT?

justinfagnani:
@robdodson FYI

robdodson:
+1 !

addyosmani:
Is this still an issue? (just asking as PR requires a rebase and it has been a few months).

robdodson:
I don't think this works yet. If I go to localhost:8080 it doesn't redirect to the components/element path, it just 404s

justinfagnani:
This patch just redirects `/components/` to `/components/{package_name}`. I'm merging this, but we should add a few more redirects like you're taking about Rob.

PR #12, opened 2015-04-23
Runs bower install before starting web server

jeffposnick:
Since the presence of a `bower_components` directory is required, and since the contents of your `bower.json` tends to change during development, running a fresh `bower install` before starting up the server seems prudent.

It's a no-op that exits quickly if nothing in your `bower.json` has changed.

### DIFF
--- a/bin/polyserve
+++ b/bin/polyserve
@@ -16,5 +16,7 @@ process.title = 'polyserve';
 
 resolve('polyserve', {basedir: process.cwd()}, function(error, path) {
   var polyserve = path ? require(path) : require('..');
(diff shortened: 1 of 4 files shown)


justinfagnani:
Running `bower install` is potentially dangerous because Bower will happily write all over your disk if configured to do so in `.bowerrc`. Me destroying some work was the initial inspiration for Polyserve. If we auto-ran `bower install` then someone who wanted to convert a project from a `.bowerrc` pointed at `../` to just using Polyserve could nuke important files.

We could possibly give hints if we detect that install should be run.

PR #23, opened 2015-05-22
add -h, -b options, fix -o

atotic:
Added some options to make it more useful:

-h for help
-b for picking browser other than default
-o now defaults to index.html, it used to open "true".
-p if unspecified, now scans ports for free port starting at 8080

This makes my workflow a lot simpler, lots of polyserves everywhere....

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,10 +2,7 @@
   "name": "polyserve",
   "version": "0.4.0",
   "description": "A simple dev server for bower components",
-  "license": {
-    "type": "BSD-3-Clause",
-    "url": "http://polymer.github.io/LICENSE.txt"
-  },
+  "license": "BSD-3-Clause",
   "repository": "PolymerLabs/polyserve",
   "main": "main.js",
   "bin": {
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   ],
   "dependencies": {
     "express": "^4.8.5",
+    "find-port": "^1.0.1",
     "minimist": "^1.1.1",
     "open": "0.0.5",
     "resolve": "^1.0.0",
--- a/src/start_server.js
+++ b/src/start_server.js
@@ -13,24 +13,50 @@ var http = require('http');
 var makeApp = require('./make_app');
 var open = require('open');
 var util = require('util');
+var findPort = require('find-port');
 
-function startServer(port, openPage, componentDir, packageName) {
-  port = port || 8080;
-  console.log('Starting Polyserve on port ' + port);
+function startWithPort(options) {
+
+  options.port = options.port || 8080;
+
+  console.log('Starting Polyserve on port ' + options.port);
 
   var app = express();
-  var polyserve = makeApp(componentDir, packageName);
+  var polyserve = makeApp(options.componentDir, options.packageName);
 
   app.use('/components/', polyserve);
 
   var server = http.createServer(app);
-  server = app.listen(port);
 
-  var baseUrl = util.format('http://localhost:%d/components/%s/', port, polyserve.packageName);
+  server = app.listen(options.port);
+
+  server.on('error', function(err) {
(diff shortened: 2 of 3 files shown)


atotic:
Just had a hunch, and improved handling of -b. Now multiple -b options open multiple browsers at once, which is nice for testing.

    polyserve -o -b 'Chrome' -b 'Firefox' -b 'Safari' FTW!

src/start_server.js
```diff
@@ -13,24 +13,54 @@ var http = require('http');
...
 
   var server = http.createServer(app);
-  server = app.listen(port);
 
-  var baseUrl = util.format('http://localhost:%d/components/%s/', port, polyserve.packageName);
+  server = app.listen(options.port);
+
+  server.on('error', function(err) {
+    if (err.code === 'EADDRINUSE')
+      console.error("ERROR: Port in use", options.port, "Aborting!");
```
justinfagnani: Let's replace "Aborting!" with something like "Please choose another port, let an unused port be chosen automatically"

atotic:
Hey, ABORTING! was a huge improvement over raw error throw :)
Fixed.
PTAL

atotic:
Fixed a license while I was in there.

src/start_server.js
```diff
@@ -13,24 +13,54 @@ var http = require('http');
 var makeApp = require('./make_app');
 var open = require('open');
 var util = require('util');
+var findPort = require('find-port');
 
-function startServer(port, openPage, componentDir, packageName) {
-  port = port || 8080;
-  console.log('Starting Polyserve on port ' + port);
+function startWithPort(options) {
```
justinfagnani: I don't like this rename, port is just one of several options.

justinfagnani: Please document the properties available in options, now that we've lost the argument names.

src/start_server.js
justinfagnani: style: wrap line to 80 cols

src/start_server.js
```diff
@@ -13,24 +13,54 @@ var http = require('http');
...
+  var baseUrl = util.format('http://localhost:%d/components/%s/', options.port,
+    polyserve.packageName);
   console.log('Files in this directory are available under ' + baseUrl);
 
-  if (openPage) {
-    open(baseUrl + openPage);
+  if (options.page) {
+    var url = baseUrl + (options.page === true ? 'index.html' : options.page);
+    if (Array.isArray(options.browser)) {
+      for (var i=0; i<options.browser.length; i++)
```
justinfagnani: style: add spaces around `=` and `<`

package.json
```diff
@@ -20,6 +20,7 @@
   ],
   "dependencies": {
     "express": "^4.8.5",
+    "find-port": "^1.0.1",
```
justinfagnani: Why this change? AFAIK we always include the link to the license.

src/start_server.js
```diff
@@ -13,24 +13,54 @@ var http = require('http');
...
+        open(url, options.browser[i]);
+    }
+    else {
+      open(url, options.browser);
+    }
+  }
+}
+
+function startServer(options) {
```
justinfagnani: ok, I see the reason for the function name above. I'd put this short function above the startWithPort for readability.

src/start_server.js
atotic: that's the way I like it too, but g-w-c team has been beating me over the head about it. done,

src/start_server.js
atotic: done

src/start_server.js
atotic: done

package.json
atotic: it's the new package.json license format. No more urls in licenses. you get a warning on install without this change.

atotic:
Fixes checked in. PTAL.

PR #24, opened 2015-05-29
Add components path options, Read default path from bowerrc

ragingwind:
Read component path from bowerrc. It's useful to use custom bower components path


justinfagnani:
This was implemented independently. Thanks and sorry for not seeing this way back!

PR #28, opened 2015-07-28
Co-development of multiple seeds (add fallback to serving ../)

doug:
Add a fallback for serving from up a directory for the co-development of multiple seed elements. This way I can keep multiple seed elements as sibling folders, but each have their own bower dependencies prior to release and bower installable sub elements.


doug:
I signed it!

doug:
This is a bad patch because it doesn't handle erroring correctly with 404 paths, but the idea of developing multiple seed elements in parallel with polyserve would be useful.

PR #30, opened 2015-10-01
Use `opn` instead of `node-open`

antleblanc:
Probably because it is actively maintained


addyosmani:
LGTM and can confirm this is a suitable replacement. `opn` is much better maintained.

PR #36, opened 2015-11-09
Read bower config from correct directory

justinfagnani:


justinfagnani:
@rictic this one will be a little easier to review commit-by-commit

bin/polyserve
```diff
@@ -9,8 +9,10 @@
  * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
```
rictic: Can replace all uses of `let` with `const` in this file

src/bower_config.js
```diff
@@ -8,16 +8,17 @@
  * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
  */
 
-var path = require('path');
-var fs = require('fs');
+'use strict';
+
+let path = require('path');
+let fs = require('fs');
```
rictic: const requires seems nice for legibility

src/make_app.js
```diff
@@ -8,34 +8,40 @@
...
-var send = require('send');
-var bowerConfig = require('./bower_config');
+'use strict';
+
+let express = require('express');
+let fs = require('fs');
+let path = require('path');
+let parseUrl = require('url').parse;
+let send = require('send');
+let bowerConfig = require('./bower_config');
```
rictic: Ditto

src/bower_config.js
justinfagnani: done

bin/polyserve
justinfagnani: done

src/make_app.js
justinfagnani: done

src/start_server.js
```diff
@@ -41,36 +52,48 @@ function startWithPort(options) {
...
+    let url = baseUrl + (options.page === true ? 'index.html' : options.page);
     if (Array.isArray(options.browser)) {
-      for (var i = 0; i < options.browser.length; i++)
+      for (let i = 0; i < options.browser.length; i++)
         opn(url, options.browser[i]);
     }
     else {
       opn(url, options.browser);
     }
   }
```
rictic: Consider waiting until after the server has actually started before telling the user or starting any browsers.

rictic:
LGTM with the above comments.

src/start_server.js
justinfagnani: It should already be doing that. `app.listen` takes a callback that's called after the server starts.http://expressjs.com/4x/api.html#app.listen which references https://nodejs.org/docs/latest/api/net.html#net_server_listen_port_hostname_backlog_callback

rictic: Yeah, but here we're calling app.listen and then immediately doing a console.log() and starting browsers, right?

I'm suggesting that we should put that logic inside of `serverStartedPromise.then(() => { ... })`

justinfagnani: Ah yes, sorry didn't parse your whole comment... good point.

PR #44, opened 2015-12-04
Fixes #41: Issue 302s to redirect instead of 301s.

### LINKED ISSUE #41: Use 302 redirect instead of 301
Chrome remembers the 301 redirect from the root path `http://localhost:8080` and makes it difficult to reuse this port for other things. I think 302 it the appropriate code for this type of redirect anyway

traviskaufman:
Issuing 302s on localhost can lead to some browsers like Chrome
incorrectly issuing redirects when another app is run on the same port
at a different time. Using 302s to redirect will prevent this.

### DIFF
--- a/src/make_app.js
+++ b/src/make_app.js
@@ -38,7 +38,7 @@ function makeApp(options) {
   let app = express();
 
   app.get('/', function (req, res) {
-    res.redirect(301, `./${packageName}/`);
+    res.redirect(`./${packageName}/`);
   });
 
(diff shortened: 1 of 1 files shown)


traviskaufman:
ping! :bell:

klebba:
Any objections to this change?

justinfagnani:
I've removed the redirects. If any are added back, I'll make sure they use 307s.

Issue #58, opened 2016-03-15
ReferenceError: Promise is not defined


iblancasa:
```bash
/usr/local/lib/node_modules/polyserve/lib/start_server.js:8
    return new Promise(function (resolve, reject) {
               ^
ReferenceError: Promise is not defined
    at Object.startServer (/usr/local/lib/node_modules/polyserve/lib/start_server.js:8:16)
    at /usr/local/lib/node_modules/polyserve/bin/polyserve:40:13
    at /usr/local/lib/node_modules/polyserve/node_modules/resolve/lib/async.js:46:14
    at process (/usr/local/lib/node_modules/polyserve/node_modules/resolve/lib/async.js:173:43)
    at ondir (/usr/local/lib/node_modules/polyserve/node_modules/resolve/lib/async.js:188:17)
    at load (/usr/local/lib/node_modules/polyserve/node_modules/resolve/lib/async.js:69:43)
    at onex (/usr/local/lib/node_modules/polyserve/node_modules/resolve/lib/async.js:92:31)
    at /usr/local/lib/node_modules/polyserve/node_modules/resolve/lib/async.js:22:47
    at Object.oncomplete (fs.js:107:15)

```

Fleuv:
I've the exact same error, by following [these steps](https://www.polymer-project.org/1.0/docs/start/first-element/intro.html):

1. Download, extract and open terminal in [polymer-first-elements](https://github.com/googlecodelabs/polymer-first-elements/releases/download/v1.0/polymer-first-elements.zip).
2. installed polyserve (on ubuntu) using `npm install -g polyserve` 
3. run `polyserve`, error...

iblancasa:
@Fleuv the problem is the version of NodeJS. Until the code will be changed (I don't know if my pull request will be accepted), you will need to use a newer version. NodeJS version in Ubuntu's repositories is so old. Install manually :D

justinfagnani:
I think we're going to move to a node support policy of supporting the current LTS and newer versions. Since 4.2 has Promises, we wouldn't consider this a bug with a supported version.

Is there any reason you can't upgrade to a more recent version of node?

iblancasa:
It is the current version in Ubuntu repositories.

justinfagnani:
0.10 has so many problems beyond the lack of Promises though. It will be very difficult to maintain support especially as our dependencies, some of which require native modules, start dropping support.

iblancasa:
[...]

PR #59, opened 2016-03-15
Solving #58

iblancasa:
It is only necessary to add these lines (promise as dependence and require the package)


iblancasa:
I signed it!

justinfagnani:
Thanks for the PR @iblancasa, but I think we're only going to support the latest LTS of node going forward. See https://github.com/PolymerLabs/polyserve/issues/58

PR #63, opened 2016-04-21
Add some tests

justinfagnani:


gulpfile.js
```diff
@@ -0,0 +1,20 @@
...
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+const fs = require('fs');
+const gulp = require('gulp');
+const mocha = require('gulp-mocha');
+
+gulp.task('test', () =>
+  gulp.src('test/*_test.js', {read: false})
+      .pipe(mocha({
+        ui: 'tdd',
```
garlicnation: you could configure mocha in mocha.opts, allowing for running tests with `mocha` on the command line

garlicnation:
LGTM w/ nit.

gulpfile.js
justinfagnani: yeah... that's how I've been doing it, but lately I've been leaning towards requiring fewer global installs. Just gulp, then everything goes from there. I don't feel too strongly though.

PR #64, opened 2016-04-21
Export TypeScript declarations in package.json

justinfagnani:
Declarations can now be pulled into dependents like polytool.


garlicnation:
LGTM

PR #69, opened 2016-04-22
fix -H option

danners:
Don't know if thats the right way to fix, since i never really used typescript, but this seems to make the -H option work again.
I don't know if the other options are broken in a similar way.


danners:
signed it

justinfagnani:
Thanks!

I think there are some other fixes to be made to args too. I'm working on properly testing them now...

PR #71, opened 2016-04-22
Serve files from root directory, redirect 404s to index.html

justinfagnani:
This makes polyserve work nice with client-side routers, app-style projects, and serving build projects.

By default any file on disk will be served. URLs that don't resolve to files, like `localhost/foo`, will serve index.html (to be configurable in follow up PR). Previous polyserve `/component` URLs still work, though redirects were removed.

`polyserve build/` will serve out of the `build/` folder.


custom_typings/send.d.ts
```diff
@@ -1,6 +1,10 @@
 declare module 'send' {
-  module send {}
+  module send {
+    interface SendError extends Error {
+      status: number;
+    }
+  }
   // TODO(justinfagnani): any->Request
-  function send(req: any, path:string): any;
+  function send(req: any, path: string, option?: any): any;
```
garlicnation: option?: ServerOptions

src/start_server.ts
```diff
@@ -95,8 +132,11 @@ function startWithPort(options: ServerOptions) {
     serverStartedReject(err);
   });
 
-  let baseUrl = `http://${options.hostname}:${options.port}/components/${polyserve.packageName}/`;
-  console.log(`Files in this directory are available under ${baseUrl}`);
+  let serverUrl = `http://${options.hostname}:${options.port}/`;
```
garlicnation:
```javascript
let serverUrl = {protocol: 'http', hostname: options.hostname, port:options.port};
let baseUrl = url.parse(url.format(serverUrl));
let basePath = `components/${options.packageName}/`
```
At line 103: in the original
```javascript
baseurl.pathname = basePath + (options.open === true ? 'index.html' : options.open);
let url = url.format(baseurl);
```

custom_typings/send.d.ts
justinfagnani: These are `send`'s options. I can go though their docs and define it.

garlicnation:
LGTM with an optional enhancement.

chuckh:
Nice work @justinfagnani!  💯  🎆  🍕   Just tested with the latest PSK2 using hashes and not using them, it worked like a charm.

PR #73, opened 2016-04-23
print usage information if invalid parameter was provided

danners:


src/cli.ts
```diff
@@ -32,14 +41,19 @@ export function run(): Promise<void> {
     }
 
     if (cliOptions.help) {
-      var usage = cli.getUsage({
-        header: 'A development server for Polymer projects',
-        title: 'polyserve',
-      });
-      console.log(usage);
+      printUsage(cli);
+      reject();
```
justinfagnani: I'd rather resolve than reject(). Rejecting a Promise is like throwing an exception, but this isn't an exceptional case, it's expected and valid user input.

src/cli.ts
```diff
@@ -21,7 +21,16 @@ export function run(): Promise<void> {
       type: Boolean,
     });
     var cli = commandLineArgs(argsWithHelp);
-    var cliOptions = cli.parse();
+
+    try {
+      var cliOptions = cli.parse();
```
justinfagnani: what case was causing `parse` to throw? I'm seeing that the parsing is rather lenient?

justinfagnani:
Thanks for the PR! I have a question about what triggered the case you're seeing. I still think this is a good idea, but would want to see a failing test to make sure the code is doing what we think.

src/cli.ts
danners: Hey, okay my line of thought was that the function should only resolve when it "runs" the server, but resolving with null could also be used to signal no server was started.

src/cli.ts
danners: The parser doesn't fail at parsing supplied options. It throws a exception when it is provided with a parameter that its not configured to parse 
See: https://www.npmjs.com/package/command-line-args#cliparseargv--object
E.g. polyserve --bla value1 would cause polyserve with this PR applied to print out the usage information. The printing of usage if a parameter can not be understood is normal behaviour for unix programs.and i think it would be nice to have.

justinfagnani: `polyserve --bla value1` isn't causing any exceptions w/o this patch. Can you add a failing test first?

src/cli.ts
justinfagnani: I think of it in the synchronous version:

```
/**
  * @options.help prints usage and throws
  * @returns a Server
  * @throws if option.help is true
 */
run(options) {}
```

We probably wouldn't this, and returning null might be ok. If you don't hate null too much.

src/cli.ts
danners: It does cause an exception, however it it catched in bin/polyserve (line 23). You can check the return value of a polyserve run with an unknown parameter, its 69 if there was a catched exception.
As for the test, i have to look into how to test this first.

danners:
version 2 in #73

PR #81, opened 2016-05-02
Cleanup weird behavior around open argument, make behavior explicit

FredKSchott:
CR @justinfagnani

### DIFF
--- a/src/args.ts
+++ b/src/args.ts
@@ -53,13 +53,20 @@ export let args : ArgDescriptor[] = [
     name: 'open',
     alias: 'o',
     description: 'The page to open in the default browser on startup.',
-    type: String,
+    type: Boolean,
   },
   {
     name: 'browser',
     alias: 'b',
-    description: 'The browser to open when using the --open option.' +
-        ' Defaults to chrome',
+    description: 'The browser(s) to open with when using the --open option.' +
+      ' Defaults to your default web browser.',
+    type: String,
+    multiple: true,
+  },
+  {
+    name: 'open-path',
+    description: 'The URL path to open when using the --open option.' +
+      ' Defaults to "index.html".',
     type: String,
   },
 ];
--- a/src/start_server.ts
+++ b/src/start_server.ts
@@ -31,11 +31,14 @@ export interface ServerOptions {
   /** The hostname to serve from */
   hostname?: string;
 
-  /** The page to open in the default browser on startup **/
-  open?: boolean | string;
+  /** Whether to open the browser when run **/
+  open?: boolean;
 
-  /** The browser to open **/
-  browser?: string | string[];
+  /** The browser(s) to open when run with open argument **/
+  browser?: string[];
+
+  /** The URL path to open in each browser **/
+  openPath?: string;
 
   /** The component directory to use **/
   componentDir?: string;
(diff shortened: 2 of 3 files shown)


src/start_server.ts
```diff
@@ -147,13 +167,15 @@ function startWithPort(userOptions: ServerOptions) {
...
-    let browsers = Array.isArray(options.browser)
-      ? <Array<string>>options.browser
-      : [options.browser];
-    browsers.forEach((browser) => {
-      opn(url.format(componentUrl), {app: browser});
-    });
+    let openUrl: url.Url = Object.assign({}, componentUrl);
+    openUrl.pathname += options.openPath;
+    if (!Array.isArray(options.openBrowser)) {
+    openWebPage(url.format(openUrl));
```
justinfagnani: style: indent 2 spaces

src/start_server.ts
```diff
@@ -147,13 +167,15 @@ function startWithPort(userOptions: ServerOptions) {
...
+    } else {
+      options.openBrowser.forEach((browser) => {
+      openWebPage(url.format(openUrl), browser);
```
justinfagnani: style: indent 2 spaces

src/start_server.ts
```diff
@@ -50,6 +53,7 @@ function applyDefaultOptions(options: ServerOptions): ServerOptions {
     port: options.port || 8080,
     hostname: options.hostname || "localhost",
     root: path.resolve(options.root || '.'),
+    openPath: options.openPath || "index.html",
```
justinfagnani: I wonder if we should default to just `/`

src/args.ts
```diff
@@ -53,13 +53,19 @@ export let args : ArgDescriptor[] = [
     name: 'open',
     alias: 'o',
     description: 'The page to open in the default browser on startup.',
+    type: Boolean,
+  },
+  {
+    name: 'open-browser',
```
justinfagnani: How do you feel about the length of these option names?

src/start_server.ts
FredKSchott: arg, Sublime is being weird for me. Good catch.

src/start_server.ts
FredKSchott: I'm fine with that, both do the same thing.

src/args.ts
FredKSchott: as opposed to just `browser` & `path`? I always prefer shorter if possible but I value the clarity (that these are open-related options and do nothing on their own) here a bit more. Happy to change them if you strongly disagree.

justinfagnani: How about `browser` and `open-path`?

FredKSchott: done

src/start_server.ts
```diff
@@ -113,6 +116,22 @@ export function getApp(options: ServerOptions) {
...
+ * Open the given web page URL. If no browser keyword is provided, `opn` will use
+ * the user's default browser.
+ */
+function openWebPage(url: string, withBrowser?: string) {
+  let openOptions = {
+    app: withBrowser
+  };
+  opn(url, openOptions, (err) => {
+    if (err) {
+      console.log(`Error launching "${openOptions.app || 'default web browser'}".`);
```
justinfagnani: I wonder if we want to return the error, rather than throw, so that it can be collected in the loop below and thrown after trying to launch all browsers. That way we don't block subsequent browsers from opening if one fails. This, of course, is totally minor relative to fixing the behavior here, and could be done later.

justinfagnani:
LGTM!

src/start_server.ts
FredKSchott: Yea I torn on this too, and don't really like how either one looks code-wise. 

As the user I'd rather be told immediately that my command wouldn't run as expected and why so that I can fix it vs. swallowing or logging it. Although if the command is wrapped/abstracted over by another tool and the user doesn't call this themselves, then if something breaks they can't fix it. That alone might be a good reason to log & continue.

FredKSchott: I'm going to change it to log the error and continue, that seems to serve more people/use cases

PR #83, opened 2016-05-02
Options parse

FredKSchott:
(Relies on #81, which has been pulled into the PR because github is bad)

This will allow polytool to better consume polyserve without having to duplicate the CLI argument mapping logic.

CR @justinfagnani


FredKSchott:
Actually now that I look at this, it might be going the wrong way (towards more of a CLI, instead of towards more of a library or even just getting pulled into polytool). Let me think about this a little more...

PR #92, opened 2016-05-17
Add debug logging

FredKSchott:
/cc @justinfagnani


src/start_server.ts
```diff
@@ -86,7 +89,7 @@ export function getApp(options: ServerOptions) {
 
   let app = express();
 
-  console.log(`Starting Polyserve...
+  logger.info(`Starting Polyserve...
```
FredKSchott: Our logs come with an `info:` prefix, what do you think about removing that prefix from the `info` logs only, so that they look more like console logs? I should be able to configure plylog to do this.

PR #95, opened 2016-05-17
Make openPath override the component path

justinfagnani:


azakus:
LGTM

FredKSchott:
Uh oh, this no longer works for elements. `-o` will now open `http://localhost:8080/index.html`, which is a blank page. It used to open `http://localhost:8080/components/element-name/index.html`, which is the correct URL.

FredKSchott:
err, wait. `-o` now does the right thing, ignore me :)

justinfagnani:
Yeah, all this does is prefer openPath to componentUrl, rather than append.

FredKSchott:
What's changed is that you now need the full component path, instead of just `index.html`. That'll be an unexpected change for anyone/anything currently using `open-path`, but one I can live with.

justinfagnani:
That's true, but polyserve hasn't been prioritizing serving application projects until now, so everything was geared to serving reusable components. We need to iterate here to serve both styles.

FredKSchott:
agreed

PR #97, opened 2016-06-01
Allow dotfiles request

nachoab:
I need to request dotfiles from my app. Would you consider using the allow option? is there a reason not to?

### DIFF
--- a/src/make_app.ts
+++ b/src/make_app.ts
@@ -67,7 +67,7 @@ export function makeApp(options: AppOptions): PolyserveApplication {
         (<any>res).append(header, headers[header]);
       }
     }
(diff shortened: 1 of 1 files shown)


nachoab:
I signed it!

justinfagnani:
Many static file servers default to not allowing dotfiles as a security measure. It's been too common that someone includes some sensitive configuration, or the .git folder, etc. How would you feel if this were behind a flag?

nachoab:
Polyserve is not for production but even then I agree that behind a flag is the way to go here and it works for me

justinfagnani:
I'm going to close this for now. If you'd like to submit a PR that adds a flag I'm all for it. Thanks!

PR #98, opened 2016-06-17
Add support for http2

tony19:
Inspired by Surma's [http2 go server](https://github.com/GoogleChrome/simplehttp2server), I thought it would be cool to test Polymer apps with http2 (although this PR uses a Node solution). I'm hoping to get this PR in so that it could be rolled into `polymer-cli`.

Note that h2 requires ALPN, which is only supported in Node 5 or newer. The server will not start on an older Node version.

This PR adds a few CLI flags for h2. If the TLS key and certificate are unspecified when using https/h2, they're automatically generated upon starting the server.

```
$ polyserve --help

  A development server for Polymer projects

polyserve

Options

...
  -P, --protocol string        The server protocol to use {h2, https/1.1,
                               http/1.1}. Defaults to "http/1.1".
  --key string                 Path to TLS certificate private key file for
                               https. Defaults to "key.pem".
  --cert string                Path to TLS certificate file for https. Defaults
                               to "cert.pem"
  --manifest string            Path to h2-push manifest
  --help                       Shows this help message
```

### Example usage
Start server with http (1.1) (default):

    $ polyserve -c .

Start server with https (1.1):

    $ polyserve -c . -P https/1.1

Start server with http2:

    $ polyserve -c . -P h2

Start http2 server with push manifest:

    $ polyserve -c . -P h2 --manifest push_manifest.json

Start https server with specific TLS certificate and key:

    $ polyserve -c . -P https/1.1 --key /path/to/key.pem --cert /path/to/cert.pem

### Verifying HTTP2
 1. Open Chrome's *DevTools > Network tab*
 2. Right-click a column header, and enable the *Protocol* column.
 3. CD into root directory of Polymer app source
 4. Run: `polyserve -c . -P h2 -o`
 5. Ignore Chrome's warning ("Your connection is not private"), caused by the self-signed certificate, and proceed to localhost.
 6. In *DevTools > Network tab*, observe the *Protocol* column shows "h2".
  screen shot 2016-06-17 at 8 04 12 am

### Verifying HTTP2 push
 1. Create a push-manifest (multi-file format). For example, use [http2-push-manifest](https://github.com/GoogleChrome/http2-push-manifest) to generate a manifest for Polymer CLI's shop-app template:

         $ cd /path/to/shop-app-template/
         $ http2-push-manifest -f index.html -f index.html

   Specifying `-f index.html` twice "tricks" it into using the multi-file format for that same file. The end result will list many files, including Bower dependencies. In my localhost tests, I noticed pushing many files actually slows down page-load (needs more performance testing).

 2. Edit `push_manifest.json` so that only a few files are pushed. Also replace *"index.html"* with *"/src/shop-app.html"*.

 For example:

 ```json
{
  "/src/shop-app.html": {
    "/src/shop-image.html": {
      "weight": 1,
      "type": "document"
    },
    "/src/shop-button.html": {
      "weight": 1,
      "type": "document"
    },
    "/src/shop-home.html": {
      "weight": 1,
      "type": "document"
    }
}
```

 3. Run: `polyserve -c . -P h2 --manifest push_manifest.json -o`

 4. Open Chrome's *DevTools > Network tab*, and refresh the Shop App's home page.

 5. In *DevTools > Network tab*, filter for `shop-` files, and verify the *Initiator* column indicates `Push/Other` for the files specified in the push-manifest.

 screen shot 2016-09-29 at 6 09 34 am

### CODE REVIEW

**Comment by TimvdLippe** [comment]
This pullrequest has been opened quite a while ago, but I really would like this feature integrated in the Polymer CLI. Currently we are shipping for HTTP2 and have to do our performance testing on our live server (with prefix, to not test on production :wink: ). Would be a lot better to have it integrated directly :)

File: src/start_server.ts [hunk 1]
```diff
@@ -181,6 +161,179 @@ function startWithPort(userOptions: ServerOptions) {
...
+}
+
+/**
+ * Writes a file asynchronously and ignores errors
+ * @param {string} filename Name of file to write
+ * @param {string} contents Desired string contents of file
+ * @returns {Promise<T>}
+ */
+function writeFile(filename: string, contents: any) {
+  let p = new Promise((resolve, reject) => {
```
justinfagnani: just directly return this Promise

src/start_server.ts
```diff
@@ -181,6 +161,179 @@ function startWithPort(userOptions: ServerOptions) {
...
+  return p;
+}
+
+/**
+ * Reads a file asynchronously
+ * @param {string} filename Name of file to read
+ * @returns {Promise<string>} Promise of file contents
+ */
+function readFile(filename: string) {
+  let p = new Promise<string>((resolve, reject) => {
```
justinfagnani: just directly return this Promise

src/start_server.ts
```diff
@@ -181,6 +161,179 @@ function startWithPort(userOptions: ServerOptions) {
...
+ * @returns {Promise<{}>} Promise of {serviceKey: string, certificate: string}
+ */
+function createTLSCertificate() {
+  return new Promise<{}>((resolve, reject) => {
+    console.log('Generating TLS certificate...');
+    pem.createCertificate({days: 1, selfSigned: true}, function(err: any, keys: any) {
+      if (err) {
+        reject(err);
+      } else {
+        writeFile('.cert.pem', keys.certificate);
```
justinfagnani: You need to wait for these Promises to complete: `Promise.all(writeFile(...), writeFile(...)).then((_) => resolve(keys));`

src/start_server.ts
```diff
@@ -181,6 +161,179 @@ function startWithPort(userOptions: ServerOptions) {
...
+ * @param {string} filename Name of file to write
+ * @param {string} contents Desired string contents of file
+ * @returns {Promise<T>}
+ */
+function writeFile(filename: string, contents: any) {
+  let p = new Promise((resolve, reject) => {
+    console.log('writing: ' + filename);
+    fs.writeFile(filename, contents, e => {
+      if (e) {
+        // just log error (don't reject)
```
justinfagnani: Why not reject?

src/start_server.ts
```diff
@@ -181,6 +161,179 @@ function startWithPort(userOptions: ServerOptions) {
...
+/**
+ * Gets the current TLS certificate (from current directory)
+ * or generates one if needed
+ * @returns {Promise<{}>} Promise of {serviceKey: string, certificate: string}
+ */
+function getTLSCertificate() {
+  let p = new Promise<{}>((resolve, reject) => {
+    let certificate: string;
+    let serviceKey: string;
+    readFile('.cert.pem')
```
justinfagnani: Can you read these in parallel?

```
let certificate: string;
let serviceKey: string;
return Promise.all([
  readFile('.cert.pem').then((d) => { certificate = d; }),
  readFile('.key.pem').then((d) => { serviceKey = d; }),
]).then((_) => {{certificate, serviceKey }})
.catch((_) => createTLSCertificate());
```

src/start_server.ts
```diff
@@ -181,6 +161,179 @@ function startWithPort(userOptions: ServerOptions) {
...
+  return p;
+}
+
+/**
+ * Gets the current TLS certificate (from current directory)
+ * or generates one if needed
+ * @returns {Promise<{}>} Promise of {serviceKey: string, certificate: string}
+ */
+function getTLSCertificate() {
+  let p = new Promise<{}>((resolve, reject) => {
```
justinfagnani: don't create a new Promise here

src/start_server.ts
```diff
@@ -181,6 +161,179 @@ function startWithPort(userOptions: ServerOptions) {
...
+}
+
+/**
+ * Creates an HTTP(S) server
+ * @param app
+ * @param {ServerOptions} options
+ * @returns {Promise<http.Server>} Promise of server
+ */
+function createServer(app: any, options: ServerOptions) {
+  return new Promise<http.Server>((resolve, reject) => {
```
justinfagnani: don't create a new Promise

src/start_server.ts
```diff
@@ -181,6 +161,179 @@ function startWithPort(userOptions: ServerOptions) {
...
+              cert: keys.certificate
+            };
+            let server = http.createServer(opt, app);
+            resolve(server);
+          })
+          .catch(reject);
+    } else {
+      let spdyOptions = {protocols: [protocol], plain: true, ssl: false};
+      let server = http.createServer({spdy: spdyOptions}, app);
+      resolve(server);
```
justinfagnani: `resolve(server)` => `return Promise.resolve(server)`

src/start_server.ts
```diff
@@ -181,6 +161,179 @@ function startWithPort(userOptions: ServerOptions) {
...
+/**
+ * Starts an HTTP(S) server on a specific port
+ * @param {ServerOptions} userOptions
+ * @returns {Promise<http.Server>} Promise of server
+ */
+function startWithPort(userOptions: ServerOptions) {
+  let options = applyDefaultOptions(userOptions);
+  let app = getApp(options);
+
+  let p = new Promise<http.Server>((resolve, reject) => {
```
justinfagnani: just return the promise here

justinfagnani:
Hey @tony19 just started to take a look at this. Sorry it's been so long.

I just realized that since we're compiling to ES6 we can use async functions. They'll make the promise using code much simpler.

Also, I don't think we can use destructuring in node 4. The tests pass because those code paths aren't hit yet.

src/start_server.ts
```diff
@@ -71,7 +76,8 @@ export function startServer(options: ServerOptions): Promise<http.Server> {
         resolve(options);
       });
     }
-  }).then<http.Server>((opts) => startWithPort(opts));
+  }).then<http.Server>((opts) => startWithPort(opts))
+    .catch((e) => console.error('ERROR: Server failed to start:', e));
```
justinfagnani: You need to propagate the error, so throw it after the `console.error`

src/start_server.ts
```diff
@@ -181,6 +161,179 @@ function startWithPort(userOptions: ServerOptions) {
...
+  componentUrl.pathname = `components/${options.packageName}/`;
+  return {serverUrl, componentUrl};
+}
+
+/**
+ * Handles server-ready tasks (prints URLs and opens browser)
+ * @param {ServerOptions} options
+ */
+function handleServerReady(options: ServerOptions) {
+  let {serverUrl, componentUrl} = getServerUrls(options);
```
justinfagnani: This won't work in node 4

src/start_server.ts
```diff
@@ -181,6 +161,179 @@ function startWithPort(userOptions: ServerOptions) {
...
+ * @param {ServerOptions} userOptions
+ * @returns {Promise<http.Server>} Promise of server
+ */
+function startWithPort(userOptions: ServerOptions) {
+  let options = applyDefaultOptions(userOptions);
+  let app = getApp(options);
+
+  let p = new Promise<http.Server>((resolve, reject) => {
+    createServer(app, options).then((server:any) => {
+      server.listen(options.port, options.hostname,
```
justinfagnani: Actually, I'd put the new Promise just around the server listening code, then you can remove the `catch()` below. Something like:

```
return createServer(app, options)
  .then((server) => new Promise((resolve, reject) => {
    server.listen(options.port, options.hostname, () => resolve(server));
    server.on('error', (e) => reject(e));
  })
  .then((_) => handleServerReady(options));
```

tony19:
@justinfagnani Thanks for the thorough review. I actually forgot about this PR. I'll update it to address your comments soon.

src/start_server.ts
tony19: Ok. Any particular reason why tsconfig is targeting ES6? If we switch to ES5 (which Node 4 supports), then we should be able to freely use any ES6 feature. It's trivial to rework this, but just wanted to make sure we're not unnecessarily catering to older versions of Node.

src/start_server.ts
tony19: Hmm. I don't recall. I'll need to investigate and expand this comment.

tony19:
@justinfagnani I've addressed all comments and added h2 push support. Note that h2 requires ALPN, which is only supported in Node5+.

justinfagnani:
@tony19 I just made some changes that require a rebase. Sorry! (if you can give me permissions on the branch, I'll update).

Looking at the new changes today :)

src/start_server.ts
justinfagnani: Mainly because node 4 supports nearly all the ES6 we want to use, except destructuring, and TypeScript only supports async/await when targeting ES6. Unfortunately we can't tell TypeScript to compile out just rest/spread, and we don't want to complicate things with a Closure/Babel pass after TypeScript.

custom_typings/spdy.d.ts
```diff
@@ -0,0 +1,148 @@
+// copied from node's "http2" module, which should have same API as spdy
```
justinfagnani: Any chance of trying to upstream this into DefinitelyTyped?

src/args.ts
```diff
@@ -69,4 +69,34 @@ export let args : ArgDescriptor[] = [
       ' Defaults to "index.html".',
     type: String,
   },
+  {
+    name: 'protocol',
+    alias: 'h',
```
justinfagnani: We're using `h` as an alias for `help` in the CLI. Things are easier if there aren't any clashes. How about `P`?

src/args.ts
```diff
@@ -69,4 +69,34 @@ export let args : ArgDescriptor[] = [
...
+    name: 'key',
+    alias: 'k',
+    description: 'Path to TLS certificate private key file for https.' +
+      ' Defaults to "key.pem".',
+    defaultValue: 'key.pem',
+    type: String,
+  },
+  {
+    name: 'cert',
+    alias: 't',
```
justinfagnani: given we share aliases with the CLI by default, how about leaving off alias for options with decent defaults. I'd leave off all but protocol.

src/start_server.ts
```diff
@@ -181,6 +193,244 @@ function startWithPort(userOptions: ServerOptions) {
...
+  }
+}
+
+/**
+ * Creates an HTTP(S) server
+ * @param app
+ * @param {ServerOptions} options
+ * @returns {Promise<http.Server>} Promise of server
+ */
+function createServer(app: any, options: ServerOptions): Promise<http.Server> {
```
justinfagnani: this function would look a bit nicer with async/await

src/start_server.ts
```diff
@@ -181,6 +193,244 @@ function startWithPort(userOptions: ServerOptions) {
...
+  let p: Promise<http.Server>;
+  if (isHttps(options.protocol)) {
+    p = getTLSCertificate(options.keyPath, options.certPath)
+        .then((keys) => {
+          let opt = {
```
justinfagnani: `const` for `opt` and `server`

src/start_server.ts
```diff
@@ -181,6 +193,244 @@ function startWithPort(userOptions: ServerOptions) {
...
+            spdy: {protocols: [options.protocol]},
+            key: keys.serviceKey,
+            cert: keys.certificate
+          };
+          let server = http.createServer(opt, app);
+          return Promise.resolve(server);
+        });
+  } else {
+    let spdyOptions = {protocols: [options.protocol], plain: true, ssl: false};
```
justinfagnani: `const`

src/start_server.ts
```diff
@@ -181,6 +193,244 @@ function startWithPort(userOptions: ServerOptions) {
...
+  return p;
+}
+
+/**
+ * Starts an HTTP(S) server on a specific port
+ * @param {ServerOptions} userOptions
+ * @returns {Promise<http.Server>} Promise of server
+ */
+function startWithPort(userOptions: ServerOptions): Promise<http.Server> {
+  let options = applyDefaultOptions(userOptions);
```
justinfagnani: `const`

test/start_server_test.js
```diff
@@ -61,4 +71,158 @@ suite('startServer', () => {
...
+        return assert.isRejected(startServer(_serverOptions));
+      } else {
+        return assert.becomes(_startStubServer(_serverOptions), _stubServer);
+      }
+    });
+
+    suite('node5+', () => {
+      // test() requires old-style function to use `this.skip()`,
+      // where `this` is the current test instance, so don't use
+      // arrow-function for the test callback.
```
justinfagnani: you could just change the suite definition function depending on the node version:

```javascript
const suiteDef = _nodeVersion < 5 ? suite.skip : suite;
suiteDef('node5+', () => {
  // ... 
});
```

src/start_server.ts
tony19: Ok, done

src/start_server.ts
tony19: Ok, done

custom_typings/spdy.d.ts
tony19: Will do

src/args.ts
tony19: Ok, done

src/args.ts
tony19: Ok, done

test/start_server_test.js
tony19: Ok, done

src/start_server.ts
tony19: Ok, done

src/start_server.ts
tony19: Ok, done

custom_typings/spdy.d.ts
tony19: typings for pem: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/11666

typings for node-spdy in progress...

justinfagnani: 🎉

src/start_server.ts
```diff
@@ -44,33 +53,59 @@ export interface ServerOptions {
...
-      resolve(options);
-    } else {
-      findPort(8080, 8180, (ports) => {
-        options.port = ports[0];
-        resolve(options);
+async function nextOpenPort(port: number): Promise<number> {
+  if (!port || port < 0) {
+    port = await new Promise<number>(resolve => {
+      findPort(8080, 8180, (ports: number[]) => {
+        resolve(ports[0]);
```
rictic: It'd be nice if this function threw if no ports were unavailable.

Something like:

```typescript
let ports = [8080, 8180];
if (!port || port < 0) {
  ports = [port];
}
return new Promise<number>((resolve, reject) => {
  findPort(ports, (availablePorts: number[]) => {
    if (availablePorts.length > 0) {
      resolve(ports[0]);
      return;
    }
    reject(new Error(`No available ports for http server to bind to. Tried: ${ports.join(', ')}`));
  });
});

```

src/start_server.ts
```diff
@@ -182,6 +195,257 @@ function startWithPort(userOptions: ServerOptions) {
...
+  server.listen(options.port, options.hostname, () => {
+    handleServerReady(options);
+  });
+
+  server.on('error', (err: any) => {
+    if (err.code === 'EADDRINUSE') {
+      console.error(portInUseMessage(options.port));
+    }
+    console.warn('rejecting with err', err);
+    throw new Error(err);
```
rictic: This won't reject the startWithPort returned promise, but it would be nice if it did.

src/start_server.ts
rictic: Ah, wait, you handle this below when creating the server which is the better time to handle this error. On the other hand, it looks like this returns `undefined` if no `port` option is given and neither 8080 or 8180 is available. Not sure what the http server library would do with an undefined port.

tony19: Interestingly, I think `find-port` has a bug because it still resolves with a port number if none are available. The redundant error checking in `startWithPort()` catches the issue. The [`getport` module](https://www.npmjs.com/package/getport) (which has the same API as `find-port`) correctly returns an error in a callback. We could switch if you prefer.

src/start_server.ts
tony19: Yeah, good catch.

I assumed throwing an error anywhere in async/await would reject, but it seems that doesn't apply inside an event handler. Now, I'm not sure how to use `async`/`await` with `createServer`. I'll probably have to `thenify` this section of code (revert to previous promise code). Any suggestions?

src/start_server.ts
justinfagnani: I wouldn't mind switching, but let's just add a TODO for now and get this PR in :)

src/start_server.ts
justinfagnani: Yep, good catch!

async/await still helps a little here, in that you can await a Promise created inline and rejections will be wired up correctly. Something like:

```typescript
await new Promise((resolve, reject) => {
  server.listen(options.port, options.hostname, () => {
    handleServerReady(options);
    resolve();
  });
  server.on('error', reject);
});
return server;
```

src/start_server.ts
tony19: Ok, done

src/start_server.ts
tony19: Cool, that works. Done

src/start_server.ts
tony19: >  it looks like this returns undefined if no port option is given and neither 8080 or 8180 is available.

@rictic When `port` is `undefined`, `!port === true`, so the body of the `if`-statement would run to find the next available port.

> Not sure what the http server library would do with an undefined port.

@rictic It would throw an error: `RangeError: "port" argument must be >= 0 and < 65536`

justinfagnani:
@tony19 I just tried to run this branch and I'm getting:

```
ERROR: Server failed to start: { Error: ENOENT: no such file or directory, open 'cert.pem'
    at Error (native) errno: -2, code: 'ENOENT', syscall: 'open', path: 'cert.pem' }
```

tony19:
@justinfagnani Fixed

justinfagnani:
Awesome, thanks!

PR #102, opened 2016-08-30
Throw a 404 for filetypes other than .html

TimvdLippe:
Not only `.html` will now trigger a 404, but also other filetypes. This is for example an issue when you are `polymer serve` an example which tries to load a `service-worker.js` without having a local dev version of the SW. Then `polymer serve` returns `index.html`, which as a consequence throws an error of "The script has an unsupported MIME type ('text/html')." which blocks all other scripts to load.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,16 @@ Navigate to `localhost:8080/components/my-element/demo.html`
   * `-b <browsername>` use this browser instead of default (ex: 'Google Chrome Canary')
   * `-H <hostname>` use this hostname instead of localhost
 
-## Run Tests
+## Contributing
+
+```
+git clone git@github.com:polymerlabs/polyserve.git
+cd polyserve/
+npm install
+npm run build
+```
+
+### Run Tests
 
 ```
 npm test
--- a/README.md
+++ b/README.md
@@ -7,14 +7,14 @@ from `./bower_components/`.
 
 ## Installation
 
-    npm install polyserve -g
+    $ npm install polyserve -g
 
 ## Usage
 
 ### Run `polyserve`
 
-    cd my-element/
-    polyserve
+    $ cd my-element/
+    $ polyserve
 
 ### Browse files
 
@@ -29,17 +29,12 @@ Navigate to `localhost:8080/components/my-element/demo.html`
   * `-b <browsername>` use this browser instead of default (ex: 'Google Chrome Canary')
   * `-H <hostname>` use this hostname instead of localhost
 
-## Contributing
+## Compiling from Source
 
-```
-git clone git@github.com:polymerlabs/polyserve.git
-cd polyserve/
-npm install
-npm run build
-```
+    $ npm run build
+
+You can compile and run Polyserve from source by cloning the repo from Github and then running `npm run build`. But make sure you have already run `npm install` before building.
 
 ### Run Tests
 
-```
-npm test
-```
+    $ npm test
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "start": "./bin/polyserve",
     "init": "typings install",
     "build": "if [ ! -f typings/main.d.ts ]; then npm run init; fi; tsc",
-    "test": "gulp test; exit 0"
+    "test": "gulp test"
   },
   "keywords": [
     "polymer",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "start": "./bin/polyserve",
     "init": "typings install",
     "build": "if [ ! -f typings/main.d.ts ]; then npm run init; fi; tsc",
-    "test": "gulp test"
+    "test": "gulp test; exit 0"
   },
   "keywords": [
     "polymer",
--- a/src/start_server.ts
+++ b/src/start_server.ts
@@ -100,17 +100,13 @@ export function getApp(options: ServerOptions) {
 
   app.use('/components/', polyserve);
 
-  const isFileRequest = (filePath: String) => {
-    return ['html', 'js', 'json', 'css', 'png', 'jpg', 'jpeg', 'gif'].filter((ext) => {
-      return filePath.endsWith(ext);
-    }).length > 0;
-  };
+  const filePathRegex: RegExp = /.*\/.+\..{1,}$/;
 
   app.get('/*', (req, res) => {
     let filePath = req.path;
     send(req, filePath, {root: root,})
       .on('error', (error: send.SendError) => {
-        if ((error).status == 404 && !isFileRequest(filePath.toLowerCase())) {
+        if ((error).status == 404 && !filePathRegex.test(filePath)) {
           send(req, '/', {root: root}).pipe(res);
         } else {
           res.statusCode = error.status || 500;
--- a/src/start_server.ts
+++ b/src/start_server.ts
@@ -98,9 +98,9 @@ export function getApp(options: ServerOptions) {
   });
   options.packageName = polyserve.packageName;
 
-  app.use('/components/', polyserve);
-
   const filePathRegex: RegExp = /.*\/.+\..{1,}$/;
+
+  app.use('/components/', polyserve);
 
   app.get('/*', (req, res) => {
     let filePath = req.path;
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -29,7 +29,16 @@ Navigate to `localhost:8080/components/my-element/demo.html`
...
   * `-H <hostname>` use this hostname instead of localhost
 
-## Run Tests
+## Contributing
+
+```
+git clone git@github.com:polymerlabs/polyserve.git
+cd polyserve/
+npm install
+npm run build
```
TimvdLippe: This was my very first project using TypeScript, so some small notes of what commands I had to execute before I was able to run it.

package.json
```diff
@@ -12,7 +12,7 @@
     "start": "./bin/polyserve",
     "init": "typings install",
     "build": "if [ ! -f typings/main.d.ts ]; then npm run init; fi; tsc",
-    "test": "gulp test"
+    "test": "gulp test; exit 0"
```
TimvdLippe: I was receiving `ELIFECYCLE` errors when running a test suite with failures and according to [this StackOverflow answer](http://stackoverflow.com/a/33534535/2761676) this was the solution. Now the error is not a terminal window full of npm errors, but instead a nicer `gulp test` error.

README.md
FredKSchott: We include this section in the polymer CLI readme:

```

## Compiling from Source

    $ npm run build

You can compile and run the CLI from source by cloning the repo from Github and then running `npm run build`. But make sure you have already run `npm install` before building.
```

Lets try to use the same copy in both READMEs. Would this have been enough info for you, as a first time TS-er?

package.json
FredKSchott: That answer you linked to mentions that `npm test` "squelches" `ELIFECYCLE` errors, and recommends only using this in non-test scripts. Were you seeing something different?

README.md
TimvdLippe: Oh yes definitely. It was not difficult to write TS, building it took me a couple of minutes to figure out by manually inspecting the `package.json`. Also updated the other command line references to the CLI format.

package.json
TimvdLippe: Without this change, the output of a failing test would be a terminal full with:

```
  17 passing (389ms)
  1 failing

  1) startServer 404s bla files:
     Error: expected 404 "Not Found", got 200 "OK"
      at Test._assertStatus (node_modules/supertest/lib/test.js:232:12)
      at Test._assertFunction (node_modules/supertest/lib/test.js:247:11)
      at Test.assert (node_modules/supertest/lib/test.js:148:18)
      at Server.assert (node_modules/supertest/lib/test.js:127:12)
      at emitCloseNT (net.js:1529:8)



[21:09:20] 'test' errored after 1.44 s
[21:09:20] Error in plugin 'gulp-mocha'
Message:
    1 test failed.

npm ERR! Linux 4.4.0-36-generic
npm ERR! argv "/usr/local/bin/node" "/usr/local/bin/npm" "run" "test"
npm ERR! node v5.9.1
npm ERR! npm  v3.10.2
npm ERR! code ELIFECYCLE
npm ERR! polyserve@0.12.0 test: `gulp test`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the polyserve@0.12.0 test script 'gulp test'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the polyserve package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     gulp test
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs polyserve
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls polyserve
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     /home/tim/Projects/polyserve/npm-debug.log
```

With this change, a failing test would be:

```
  17 passing (131ms)
  1 failing

  1) startServer 404s bla files:
     Error: expected 404 "Not Found", got 200 "OK"
      at Test._assertStatus (node_modules/supertest/lib/test.js:232:12)
      at Test._assertFunction (node_modules/supertest/lib/test.js:247:11)
      at Test.assert (node_modules/supertest/lib/test.js:148:18)
      at Server.assert (node_modules/supertest/lib/test.js:127:12)
      at emitCloseNT (net.js:1529:8)



[21:10:40] 'test' errored after 268 ms
[21:10:40] Error in plugin 'gulp-mocha'
Message:
    1 test failed.
```


I can of course revert this change if you want to, but for me the second output is a lot more readable and easier to digest what the actual problem is. Up to you :smile:

README.md
FredKSchott: Nice! big 👍 on adding the other updates

src/start_server.ts
```diff
@@ -100,11 +100,17 @@ export function getApp(options: ServerOptions) {
 
   app.use('/components/', polyserve);
 
+  const isFileRequest = (filePath: String) => {
+    return ['html', 'js', 'json', 'css', 'png', 'jpg', 'jpeg', 'gif'].filter((ext) => {
```
FredKSchott: Can we make this a bit more flexible? I don't like the idea of only supporting and maintaining a subset of all possible file types.

If we want to return a 404 for all paths that include a file extension, can you look into a solution using either regex matching or something with `lastIndexOf(".")`?

test/start_server_test.js
```diff
@@ -51,14 +51,17 @@ suite('startServer', () => {
       .end(done);
   });
 
-  test('404s .html files', (done) => {
-    let app = getApp({
-      root: __dirname,
-    });
-    supertest(app)
-      .get('/foo.html')
+  ['html', 'js', 'json', 'css', 'png', 'jpg', 'jpeg', 'gif'].forEach((ext) => {
```
FredKSchott: love using this list as a test dataset though! If we can make the above more flexible, you could remove a few of these and/or add in some less common ones.

src/start_server.ts
TimvdLippe: Yes I was thinking of regex match, but hesitated because of the "e-mail regex" problem. We might catch too many, too few, what is the length of the file extensions etc...

A basic one would be `filePath.matches(/.+\.\w+$/)`. This could potentially catch too much though, I am not 100% up-to-date with the specification of a file URL.

package.json
FredKSchott: Hmm, I can't reproduce. How are you running these tests? If you're running `npm run test`, can you try running `npm test` instead and posting the output? I think that's required for npm to swallow those errors automatically.

TimvdLippe: Yes you are completely correct. TIL. Reverting this change :)

src/start_server.ts
FredKSchott: Yea I agree it's not perfect either, but I like it better than the explicit list. Lets keep it simple at first and match all url paths where an extension (period & 1+ other characters) exists after the last slash.

package.json
FredKSchott: Nice!

knowledge is power!

src/start_server.ts
TimvdLippe: Done :tada:

FredKSchott:
Tested the regex on shop directory, everything looked good to me: https://regex101.com/r/vK7sZ5/1

Obviously the one case that this won't catch is when files don't have extensions (ex: LICENSE) but that should be rare and I can't say I've ever seen that in the wild. We can look into using `fs.stat()` or something similar to analyze the file/folder if this somehow ends up becoming an issue.

src/start_server.ts
```diff
@@ -100,11 +100,13 @@ export function getApp(options: ServerOptions) {
 
   app.use('/components/', polyserve);
 
+  const filePathRegex: RegExp = /.*\/.+\..{1,}$/;
```
FredKSchott: nit: can you move this above ~L86 or below ~L109? Stylistically it feels odd here in between our two route definitions.

FredKSchott:
LGTM with 1 nit comment! Ready to merge once that's addressed

src/start_server.ts
TimvdLippe: Done :+1:

FredKSchott:
Great, thanks @TimvdLippe!

PR #103, opened 2016-09-03
Check if bower.json exists to prevent incorrect console error

TimvdLippe:
Before calling `bowerConfig(root)`, check if the file actually exists.

Fixes https://github.com/Polymer/polymer-cli/issues/357

### DIFF
--- a/src/bower_config.ts
+++ b/src/bower_config.ts
@@ -13,11 +13,7 @@ import * as fs from 'fs';
 
 function bowerConfigPath(root?: string): string {
   root = root || process.cwd();
-  try {
-    return path.resolve(root, 'bower.json');
-  } catch (e) {
-    return null;
-  }
+  return path.resolve(root, 'bower.json');
 }
 
 function bowerConfigContents(root?: string): string {
@@ -33,11 +29,9 @@ function bowerConfigContents(root?: string): string {
 }
 
 export function bowerConfig(root?: string) {
+  const config = bowerConfigContents(root);
   try {
-    let config = bowerConfigContents(root);
-    if (config) {
-      return JSON.parse(config);
-    }
+    return JSON.parse(config);
   } catch (e) {
     console.error('Could not parse bower.json');
     console.error(e);
--- a/src/bower_config.ts
+++ b/src/bower_config.ts
@@ -13,7 +13,11 @@ import * as fs from 'fs';
 
 function bowerConfigPath(root?: string): string {
   root = root || process.cwd();
-  return path.resolve(root, 'bower.json');
+  try {
+    return path.resolve(root, 'bower.json');
+  } catch (e) {
+    return null;
+  }
 }
 
 function bowerConfigContents(root?: string): string {
@@ -22,8 +26,7 @@ function bowerConfigContents(root?: string): string {
   try {
     contents = fs.readFileSync(bowerConfigPath(root)).toString();
   } catch (e) {
-    console.error('Error reading config at ' + bowerConfigPath());
-    console.error(e);
+    return '{}'
   }
 
   return contents || '{}';
@@ -31,7 +34,10 @@ function bowerConfigContents(root?: string): string {
 
 export function bowerConfig(root?: string) {
   try {
-    return JSON.parse(bowerConfigContents(root));
+    let config = bowerConfigContents(root);
+    if (config) {
+      return JSON.parse(config);
+    }
   } catch (e) {
     console.error('Could not parse bower.json');
     console.error(e);
--- a/src/make_app.ts
+++ b/src/make_app.ts
@@ -46,8 +46,8 @@ export function makeApp(options: AppOptions): PolyserveApplication {
     packageName = bowerConfig(root).name;
   }
   if (!packageName) {
-    console.log('bower.json not found, falling back to current path as package name');
-    packageName = path.basename(process.cwd());
+    packageName = path.basename(root || process.cwd());
+    console.log(`no bower.json detected, using name "${packageName}"`);
   }
   let headers = options.headers || {};
 
--- a/src/make_app.ts
+++ b/src/make_app.ts
@@ -41,9 +41,15 @@ export function makeApp(options: AppOptions): PolyserveApplication {
   options = options || {};
   const root = options.root;
   const componentDir = options.componentDir || 'bower_components';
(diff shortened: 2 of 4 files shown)


FredKSchott:
@TimvdLippe this is great! We've been talking about doing this forever now but have never had time to actually tackle it. 

Instead of adding a check for the file before the call to `bowerConfig()`, can we instead have `bowerConfig()` return null if an error occurs finding & parsing `bower.json`? Then we would just check for null and fall back to `path.basename(process.cwd())` if it fails.

TimvdLippe:
Thank you! Returning would be falsy, but if you would change the object to null you dont suppress the error. Do you mean that the console.error should be removed and replaced by return null? In that case sometimes when you would expect an error message you dont get one.

FredKSchott:
I'm suggesting something like this:

```js
if (!packageName){
  let bowerData = bowerConfig(root);
  packageName = (bowerData) ? bowerData.name : path.basename(process.cwd());
}
```

What do you think?

TimvdLippe:
@FredKSchott That code is equivalent to the current state. The problem is that deep down `bowerConfig` errors to the `console` that it could not find any config and then coincidentally `bowerConfig(root).name` is undefined and the fallback of `path.basename(process.cwd())` is used.

So to suppress the error we either have to remove it, or (like the current state of the PR) pre-emptively check if the file actually exists to prevent the console error.

FredKSchott:
I'm confused how that's possible, it looks like the entirety of `bowerConfig` is wrapped in a try-catch: https://github.com/TimvdLippe/polyserve/blob/e04cb9c3ad66d737f4e15a513277e0be4f9958e6/src/bower_config.ts#L32

Is there something I'm missing? Or are we just talking about the `console.error()` output?

TimvdLippe:
@FredKSchott Yes that is correct, but the try-catch does write to the console if it encountered an error. I am just talking about the console error output. The actual console error described in https://github.com/Polymer/polymer-cli/issues/357 is the one in `bowerConfigContents(root)`: https://github.com/TimvdLippe/polyserve/blob/e04cb9c3ad66d737f4e15a513277e0be4f9958e6/src/bower_config.ts#L22-L27

E.g. if the file can not be found, write the error to console. However, since https://github.com/PolymerLabs/polyserve/blob/master/src/make_app.ts#L43 means that if no bower.json is found, fallback to `path.basename(process.cwd())`, both an error is written to console AND we continue as expected. Therefore either we have to prevent the error from happening, or not write the error at all in `bowerConfigContents(root)`

FredKSchott:
Ah, got it. I think the problem is that we log an error within those helper functions, instead of propagating the error (or null, or some other falsey response) and then letting polyserve decide whether it is worth logging.

Could we remove the error logging inside of `bowerConfigContents()` & `bowerConfig()`, and instead return an error or null? Then the caller would decide what it wants to do.

FredKSchott:
For context, the thing that we've been talking about which I mentioned in my first comment is our desire to support serving build targets (`/build/bundled`) without needing to include the `bower.json` file in the build itself. Right now they log that error unless bower.json is included. Polymer serve should at most log a debug message saying that bower.json wasn't found and that it's continuing with `path.basename(process.cwd())`.

TimvdLippe:
I have now removed the error log and added a debug message in `makeApp` to notify the fallback to the current path as package name.

TimvdLippe:
@FredKSchott I think this PR can be merged and a new release of polyserve can be published? :smile:

src/bower_config.ts
```diff
@@ -13,7 +13,11 @@ import * as fs from 'fs';
 
 function bowerConfigPath(root?: string): string {
   root = root || process.cwd();
-  return path.resolve(root, 'bower.json');
+  try {
+    return path.resolve(root, 'bower.json');
```
FredKSchott: `path.resolve()` is only responsible for resolving file path fragments. It will never throw (see https://nodejs.org/docs/latest/api/path.html#path_path_resolve_path)

src/bower_config.ts
```diff
@@ -22,16 +26,18 @@ function bowerConfigContents(root?: string): string {
...
+    return '{}'
   }
 
   return contents || '{}';
 }
 
 export function bowerConfig(root?: string) {
   try {
-    return JSON.parse(bowerConfigContents(root));
+    let config = bowerConfigContents(root);
```
FredKSchott: Since this will never throw, can you move it out of the try-catch block?

src/bower_config.ts
```diff
@@ -22,16 +26,18 @@ function bowerConfigContents(root?: string): string {
...
+    if (config) {
```
FredKSchott: bowerConfigContents will always be a string, no need for check here

src/make_app.ts
```diff
@@ -40,8 +40,14 @@ export function makeApp(options: AppOptions): PolyserveApplication {
...
   let root = options.root;
   let componentDir = options.componentDir || 'bower_components';
-  let packageName = options.packageName || bowerConfig(root).name
-      || path.basename(process.cwd());
+  let packageName = options.packageName;
+  if (!packageName){
+    packageName = bowerConfig(root).name;
+  }
+  if (!packageName) {
+    console.log('bower.json not found, falling back to current path as package name');
```
FredKSchott: Can we reword this to sound less error-like? We want this to work for application builds as well as elements.

How about:

```js
packageName = path.basename(process.cwd());
console.log(`no bower.json detected, using name "${packageName}"`);
```

test/make_app_test.js
```diff
@@ -14,6 +25,18 @@ suite('makeApp', () => {
...
+  test('shows friendly error when bower.json does not exist', () => {
+    let called = false;
+    console.error = function(e) {
+      called = true;
+    }
+    let app = makeApp({
+      root: resolve(__dirname, 'no_bower_json/')
+    });
+    assert.isFalse(called);
+    assert.equal(app.packageName, 'polyserve');
```
FredKSchott: isn't packageName supposed to be `no_bower_json` in this case?

TimvdLippe:
Processed feedback and got tests working :tada:

TimvdLippe:
@FredKSchott Ping :smile:

FredKSchott:
Thanks for the awesome work @TimvdLippe!

PR #104, opened 2016-09-30
0.13

justinfagnani:
Upgrade command-line-args to 3.0.1
Upgrade to TypeScript 2.0
Small updates to README
Constant gardening
Use import/require for non-ES modules
Remove ‘use strict’;
v0.13


package.json
```diff
@@ -10,29 +10,33 @@
...
+    "@types/serve-static": "^1.7.31",
+    "command-line-args": "^3.0.1",
+    "command-line-usage": "^3.0.3",
     "express": "^4.8.5",
     "find-port": "^1.0.1",
     "opn": "^3.0.2",
     "resolve": "^1.0.0",
-    "send": "^0.10.1"
+    "send": "^0.10.1",
+    "typescript": "^2.0.3"
```
rictic: tsc should be a dev dep yeah?

src/cli.ts
```diff
@@ -44,29 +39,31 @@ export function run(): Promise<void> {
...
+      printUsage(argsWithHelp);
     } else if (cliOptions.version) {
       console.log(getVersion());
-      resolve(null);
     } else {
-      return startServer(options);
+      await startServer(options);
     }
-  });
+  } catch (e) {
```
rictic: we probably don't want to catch errors with starting the server here

test/cli_test.js
```diff
@@ -15,19 +15,16 @@ const cliRun = require('../lib/cli').run;
...
 
-    result.then(function() {
-      process.argv = argv;
-      done();
-    }, function(err) {
-      process.argv = argv;
-      done(err);
-    });
+    try {
+      return cliRun();
```
rictic: I think what you want here is:

`return await cliRun();`

With just returning `cliRun` directly you're resetting the argv before the first microtask, which probably works for now but is fragile

justinfagnani: The test isn't `.ts` so no `await`. This works because all the work in `cliRun` is sync, except for starting the server. It's fragile, but I'll fix when we move tests to TS.

package.json
justinfagnani: oops

test/cli_test.js
rictic: Arg. Ok. TODO?

PR #105, opened 2016-10-08
Proxy specified path to different host and port

TimvdLippe:
Today I was converting an old PSK1 based application to PSK2. In that application we used `browserSync` to serve all the files. I wanted to get rid of the old configuration and adopt `polyserve`, but discovered that we were using proxies for our backend. E.g. we want to redirect all requests starting with `/v1/` to a different port (where our backend is running).

This simple change allows to redirect requests to such a port. Usage of the command line is

```
polyserve --root build/bundled/ --proxy-root v1 --proxy-target localhost --proxy-port 9000
```


TimvdLippe:
@FredKSchott so I have been trying to make the build pass and write a test for this, but have been unsuccesful thus far :sob: 

First of all `npm run build` fails on the CI, but I have been unable to replicate this issue. It is building just fine for me locally.
Secondly I have been trying to set up a proxy server and redirect requests in a test. While it works for me by using the command line, I was unable to create the proxy server with `spdy` and make it work.I am probably missing some basic information about node and http servers.

Other than that I removed the `proxy-port`, since it felt redundant. You would never use just `port`, so making it 1 argument for the proxy target felt better.

TimvdLippe:
Seems the build is failing due to #107

custom_typings/proxy.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module 'proxy' {
```
FredKSchott: what is this custom_typing supposed to be? I don't see a `proxy` package in our package.json.

src/start_server.ts
```diff
@@ -143,6 +150,24 @@ export function getApp(options: ServerOptions): express.Express {
 
   app.use('/components/', polyserve);
 
+  if (options.proxyRoot) {
```
FredKSchott: `options.proxyRoot && options.proxyTarget`

src/start_server.ts
```diff
@@ -143,6 +150,24 @@ export function getApp(options: ServerOptions): express.Express {
...
+    const httpProxy = require('http-proxy');
+    const apiProxy = httpProxy.createProxyServer();
+    const proxy = (req: express.Request, res: express.Response) => {
```
FredKSchott: overloading the `proxy` import, I'm surprised we don't have linting to catch small errors like this

src/start_server.ts
```diff
@@ -143,6 +150,24 @@ export function getApp(options: ServerOptions): express.Express {
 
   app.use('/components/', polyserve);
 
+  if (options.proxyRoot) {
+    const httpProxy = require('http-proxy');
```
FredKSchott: Put all of our module loading at the top of the file. You can do `import httpProxy = require('http-proxy')` in TS.

src/start_server.ts
```diff
@@ -143,6 +150,24 @@ export function getApp(options: ServerOptions): express.Express {
...
+    const apiProxy = httpProxy.createProxyServer();
+    const proxy = (req: express.Request, res: express.Response) => {
+      // Remove proxied part from the url
+      req.url = req.url.slice(options.proxyRoot.length + 1);
+      apiProxy.web(req, res, { target: options.proxyTarget }, (e: proxy.ProxyError) => {
```
FredKSchott: nit: `e` => `err`, `error`, `proxyError`, etc.

test/start_server_test.js
```diff
@@ -74,6 +74,28 @@ suite('startServer', () => {
...
+        port: 8080
+      }).then(() => {
+        startServer({
+          port: 5375,
+          root: __dirname,
+          proxyRoot: 'normally-non-existing-path',
+          proxyTarget: 'http://localhost:8080'
+        }).then((app) => {
+          supertest(app)
+          .get('/normally-non-existing-path/bower_components/test-file.txt')
```
FredKSchott: this path is a little strange, can you just do `/normally-non-existing-path/test-file.txt` or `/normally-non-existing-path/test/test-file.txt`?

test/start_server_test.js
```diff
@@ -74,6 +74,28 @@ suite('startServer', () => {
...
+          .expect(200, 'Proxy response')
```
FredKSchott: where does this come from?

src/args.ts
```diff
@@ -96,4 +96,14 @@ export let args : ArgDescriptor[] = [
     description: 'Path to h2-push manifest',
     type: String,
   },
+  {
+    name: 'proxy-root',
+    description: 'Url to proxy from',
```
FredKSchott: A little more detail here: full URL or a path?

src/args.ts
```diff
@@ -96,4 +96,14 @@ export let args : ArgDescriptor[] = [
...
+    type: String,
+  },
+  {
+    name: 'proxy-target',
+    description: 'Url to proxy to',
```
FredKSchott: A little more detail here: full URL or a path? reference that this is related to proxy-target?

TimvdLippe:
@FredKSchott Thanks for the review, the comments have been fixed! I also fixed the tests, it was a very stupid error: I forgot the `test-component/` folder in the request URL.

src/start_server.ts
```diff
@@ -17,13 +17,16 @@ import * as http from 'spdy';
...
 import * as send from 'send';
+import * as proxy from 'http-proxy';
 import * as url from 'url';
 import * as fs from 'mz/fs';
 
 import { makeApp } from './make_app';
 
 import findPort = require('find-port');
 import opn = require('opn');
+// Cant be an import since httpProxy is not a function, but an object instead
```
FredKSchott: no need to comment on this. Or, if you do want to add a comment, add it above these three since it's the reason all 3 are using require.

FredKSchott:
Also it looks like you need to rebase/merge.

src/args.ts
```diff
@@ -96,4 +96,14 @@ export let args : ArgDescriptor[] = [
...
     type: String,
   },
+  {
+    name: 'proxy-root',
+    description: 'Top-level folder that should be redirected to the proxy-target. E.g. `foo` when you want to redirect all requests of `https://localhost/foo/`',
+    type: String,
+  },
+  {
+    name: 'proxy-target',
+    description: 'Host url to proxy to, for example `https://myredirect:8080/foo`',
```
justinfagnani: Capitalize URL.

src/args.ts
```diff
@@ -96,4 +96,14 @@ export let args : ArgDescriptor[] = [
     description: 'Path to h2-push manifest',
     type: String,
   },
+  {
+    name: 'proxy-root',
```
justinfagnani: Maybe name this `proxy-path`, since it's used in the location of the path argument of `app.get()`, etc.

src/start_server.ts
```diff
@@ -143,6 +152,23 @@ export function getApp(options: ServerOptions): express.Express {
...
+    const proxyHandler = (req: express.Request, res: express.Response) => {
+      // Remove proxied part from the url
+      req.url = req.url.slice(options.proxyRoot.length + 1);
+      apiProxy.web(req, res, { target: options.proxyTarget }, (error: proxy.ProxyError) => {
+        res.statusCode = error.status || 500;
+        res.end(error.message);
+      });
+    };
+
+    app.post(`/${options.proxyRoot}/*`, proxyHandler);
```
justinfagnani: Use `app.all()`

justinfagnani: We should either validate that the path doesn't include '*', '+' and '?', or escape them before handing to express.

src/start_server.ts
```diff
@@ -143,6 +152,23 @@ export function getApp(options: ServerOptions): express.Express {
 
   app.use('/components/', polyserve);
 
+  if (options.proxyRoot && options.proxyTarget) {
+    const apiProxy = httpProxy.createProxyServer();
+    const proxyHandler = (req: express.Request, res: express.Response) => {
+      // Remove proxied part from the url
+      req.url = req.url.slice(options.proxyRoot.length + 1);
```
justinfagnani: Is `req.path` a getter? ie, is it automatically updated when `req.url` is set?

justinfagnani: It might be good to have an assertion that the request path starts with the proxy root to guard against cases where express unexpectedly matches something else based on special characters in the proxy root.

src/start_server.ts
justinfagnani: Also validate that proxy root isn't `components`

test/start_server_test.js
```diff
@@ -74,6 +74,32 @@ suite('startServer', () => {
     })
   });
 
+  suite('proxy', () =>  {
+    test('rewrites directory with proxy', (done) => {
+      startServer({
+        root: __dirname,
+        port: 8080
```
justinfagnani: hard coding the port, especially a common one like this, could cause the tests to fail.

src/start_server.ts
TimvdLippe: If I understand [the documentation](http://expressjs.com/en/api.html#req.originalUrl) correctly, yes. Please correct me if I am wrong

TimvdLippe:
Thanks a lot for the review. I agree with the cli vs internal proxy representation and have updated the code accordingly. PR is also rebased 😄

src/start_server.ts
```diff
@@ -151,6 +156,32 @@ export function getApp(options: ServerOptions): express.Express {
 
   app.use('/components/', polyserve);
 
+  if (options.proxy) {
+    if (options.proxy.path === 'components') {
```
justinfagnani: `options.proxy.path.startsWith('components')`, and might need to do this after conditionally adding the leading `/`

src/start_server.ts
```diff
@@ -151,6 +156,32 @@ export function getApp(options: ServerOptions): express.Express {
...
+      console.error('proxy path can not start with components.');
+      return;
+    }
+
+    const proxyPath = options.proxy.path
+                             .replace('*', '\*')
```
justinfagnani: There's two problems here. 1) the first argument to replace() will match even escaped '*'s 2) the right hand side needs to escape the backslash.

I think this is actually easier to write without regexes. This will probably work, or maybe there's a package that handles this:

```typescript
let escapedPath = '';
let escaped = false;
for (const char of options.proxy.path) {
  escaped = (char === '\\') ? !escaped : false;
  if (escaped && (char === '*' || char === '+' || char === '?')) {
    escapedPath += '\\';
  }
  escapedPath += char;
}
```

src/start_server.ts
```diff
@@ -151,6 +156,32 @@ export function getApp(options: ServerOptions): express.Express {
...
+        // Remove proxied part from the url
+        req.url = req.url.slice(proxyPath.length + 1);
+      }
+      apiProxy.web(req, res, { target: options.proxy.target }, (error: proxy.ProxyError) => {
+        res.statusCode = error.status || 500;
+        res.end(error.message);
+      });
+    };
+
+    app.all(`/${proxyPath}/*`, proxyHandler);
```
justinfagnani: Last bit of validation / conditional logic on the path: Handle cases where path starts or ends with `/`.

TimvdLippe:
I have finished writing several tests for the various escaping and path manipulation. Sadly I bumped into a problem where `supertest` succeeds tests even if they should not. E.g. `polyserve` returns a 500, whereas `supertest` expects a 200. Even though the statuscode is different, the test still passes.

Therefore the updated PR is a bit delayed until I can figure out why this is happening.

TimvdLippe:
Okay I was able to fix the tests. There are 2 failing tests left which test the regexes. For some reason the get request in the test does not match on the escapedPath and therefore does not redirect to the proxy. I also had to change the dependency from `http-proxy` to `http-proxy-middleware` which has `pathRewrite` which fixes the `req.url` rewriting

src/start_server.ts
```diff
@@ -150,7 +155,40 @@ export function getApp(options: ServerOptions): express.Express {
...
+      console.error('proxy path can not start with components.');
+      return;
+    }
+
+    let escapedPath = '';
+    let escaped = false;
+    for (const char of options.proxy.path) {
+      escaped = (char === '\\') ? !escaped : false;
+      if (!escaped && (char === '*' || char === '+' || char === '?')) {
+        escapedPath += '\\\\';
```
justinfagnani: Why are you writing two backslashes here? This would cause the following character to still be unescaped, unless I'm missing something.

TimvdLippe: I had thought the same. If these are 2 backslashes, the following syntax errors are thrown when compiling the RegExp for the Express path:

```
  1) startServer proxy escapes path with regex symbols:
     Error: SyntaxError: Invalid regular expression: /^\/\+regex\(.*)path\?\/?(?=\/|$)/: Unmatched ')'
      at lib/start_server.js:77:19
      at next (native)
      at fulfilled (lib/start_server.js:17:58)

  2) startServer proxy does not unescape escape symbols in path:
     Error: SyntaxError: Invalid regular expression: /^\/\+regex\?path\(.*)\/?(?=\/|$)/: Unmatched ')'
      at lib/start_server.js:77:19
      at next (native)
      at fulfilled (lib/start_server.js:17:58)
```

It seems that express simply looks for `*` in the route and then replaces, without properly escaping. The underlying path matching of `express` uses `path-to-regexp` which has this comment in the code: https://github.com/pillarjs/path-to-regexp/blob/master/index.js#L18-L20

E.g. if you want to escape special characters, double backslashes should be used.

I have tried to use 2 backslashes for everything except the express path, but that was unsuccessful too.

TimvdLippe:
This PR just needs a rebase back onto master, then it is ready I think?

TimvdLippe:
@justinfagnani This is ready for review again :tada:

justinfagnani:
Thanks @TimvdLippe !

PR #111, opened 2016-11-17
Reference custom typings so that they are properly exported

FredKSchott:
TS consumers are not currently able to compile their projects when polyserve is included.

/cc @rictic @justinfagnani


FredKSchott:
Thanks Peter!

PR #112, opened 2016-11-17
Add tools for automatically running tests.

rictic:
This adds a vscode test task, as well as an npm script to automatically rerun tests on change.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,35 @@
+{
+  "version": "0.1.0",
+  "command": "npm",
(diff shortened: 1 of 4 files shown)


.vscode/tasks.json
```diff
@@ -0,0 +1,31 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=733558
```
FredKSchott: The current plan is no comments (here and settings)

rictic: Removed!

PR #114, opened 2016-11-17
Convert tests to typescript

rictic:


gulpfile.js
```diff
@@ -1,20 +1,23 @@
...
+const mocha = require('gulp-spawn-mocha');
 
-gulp.task('test', () =>
-  gulp.src('test/*_test.js', {read: false})
-      .pipe(mocha({
-        ui: 'tdd',
-        reporter: 'spec',
-      }))
-);
+gulp.task('test', function() {
```
FredKSchott: I don't love gulp + mocha, the need for gulp-spawn-mocha, and the unnecessary need for istanbul. If we're not doing anything special here, can we just run this via an npm build script? We're already doing this in `.vscode/`

rictic: Speaking my language. Done.

src/test/make_app_test.ts
```diff
@@ -0,0 +1,37 @@
+'use strict';
```
justinfagnani: remove

src/test/cli_test.ts
```diff
@@ -1,23 +1,27 @@
...
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
  * Code distributed by Google as part of the polymer project is also
- * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
  */
 
 'use strict';
```
justinfagnani: remove

src/test/bower_config_test.ts
```diff
@@ -0,0 +1,14 @@
+'use strict';
```
justinfagnani: remove

src/test/start_server_test.ts
```diff
@@ -0,0 +1,292 @@
...
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+'use strict';
+
+import {getApp, ServerOptions} from '../start_server';
```
justinfagnani: Did clang order this way? I think our import order would put paths below names

rictic:
merged into master

PR #115, opened 2016-11-17
Some improvements to tests and code organization

rictic:
This is a noop change that:
  * adds a test for the components_dir option
  * converts tests to use async/await
  * moves some private functions out of src/start_server.ts for modularity / legibility reasons


rictic:
Tracking down the node v6 failure now.

rictic:
Done

FredKSchott:
Love the improved tests!
Love the refactored helper methods!
Can you explain why you chose to organize them in a folder named `internal`?

justinfagnani:
Give me a bit to review this

rictic:
Yeah, as a maintainer it wasn't clear what the public API of the module is vs the private API. This was unambiguous that the internal files are def internal

FredKSchott:
for clarity, I think the `internal/` move is a good step forward but it doesn't solve the whole internal vs. external interface problem. For example, is everything exported by a non-`internal` module part of the public interface? What if I want to export a function inside of a `non-internal` package without adding it to the public interface?

Since this package already defines an explicit public interface in `polyserve.ts` I'm +1 on this PR, but if we want to support a file-system-based interface more broadly across our tooling i'd like us to discuss that more.

src/util/open_browser.ts
```diff
@@ -0,0 +1,60 @@
...
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {ServerOptions} from '../start_server';
+import opn = require('opn');
+import * as url from 'url';
+
+/**
+ * Open the given web page URL. If no browser keyword is provided, `opn` will
+ * use
```
justinfagnani: redo line breaks

src/util/open_browser.ts
```diff
@@ -0,0 +1,60 @@
...
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {ServerOptions} from '../start_server';
+import opn = require('opn');
```
justinfagnani: put import/requires last

src/util/open_browser.ts
```diff
@@ -0,0 +1,60 @@
...
+      // log error and continue
+      console.error(
+          `ERROR: Problem launching "${openOptions.app || 'default web browser'
+                                                          }".`);
+    }
+  });
+}
+
+/**
+ * Opens a browser
```
justinfagnani: can you add that this can open multiple browsers? so we can tell the difference from `openWebPage`

src/util/push.ts
```diff
@@ -0,0 +1,90 @@
+import * as fs from 'fs';
```
justinfagnani: license header

src/util/open_browser.ts
rictic: Done

src/util/open_browser.ts
rictic: Done

src/util/open_browser.ts
rictic: Done

src/util/push.ts
rictic: Done

PR #116, opened 2016-11-18
Compile JS with Babel for browsers that need it.

justinfagnani:


src/compile-middleware.ts
```diff
@@ -0,0 +1,107 @@
+/**
+ * @license
+ * Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
```
notwaldorf: 2016! Looks like a new file :)

src/make_app.ts
```diff
@@ -13,8 +13,11 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { parse as parseUrl } from 'url';
 import { bowerConfig } from './bower_config';
+import { babelCompile } from './compile-middleware';
+import {Response, Request} from 'express';
```
notwaldorf: Space around the { like in the other imports?

src/transform-middleware.ts
```diff
@@ -0,0 +1,81 @@
+/**
+ * @license
+ * Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
```
notwaldorf: 2016

src/compile-middleware.ts
```diff
@@ -0,0 +1,108 @@
...
+      if (contentType === htmlMimeType) {
+        const document = parse5.parse(body);
+        const scriptTags = dom5.queryAll(document, isInlineJavaScript);
+        for (const scriptTag of scriptTags) {
+          try {
+            const script = dom5.getTextContent(scriptTag);
+            const compiledScriptResult = babelCore.transform(script, {
+              presets: [babelLatest],
+            });
+            dom5.setTextContent(scriptTag, compiledScriptResult.code);
```
rictic: I'd recommend pulling the babelCore call into a `compile` function, just because it's highly likely that we'll want to tweak it over time, and we definitely don't want inline vs separate file compilation to drift

src/compile-middleware.ts
```diff
@@ -0,0 +1,108 @@
...
+
+const isInlineJavaScript = dom5.predicates.AND(
+  dom5.predicates.hasTagName('script'),
+  dom5.predicates.NOT(dom5.predicates.hasAttr('src')));
+
+function needCompilation(uaParser: UAParser): boolean {
+  const engine = uaParser.getEngine();
+  const browser = uaParser.getBrowser();
+  const versionSplit = browser.version && browser.version.split('.');
+  const majorVersion = versionSplit ? parseInt(versionSplit[0], 10) : null;
```
rictic: Recommend giving the fallback version a number, for clarity

e.g. it's easy to see that a version number of `-1` wouldn't match any of the below comparisons without diving into obscure ES semantics

src/make_app.ts
```diff
@@ -65,10 +70,17 @@ export function makeApp(options: AppOptions): PolyserveApplication {
...
       }
     }
-    send(req, filePath).pipe(res);
+
+    const _send = send(req, filePath, {
+      lastModified: false,
+      cacheControl: false,
+    });
+    // Uncomment this to disable disable 304s from send() and always compile
+    // _send.isFresh = () => false;
```
rictic: What's the issue here? How does this relate to the lastModified and cacheControl options just above?

Also, nit, `disable disable` -> `disable`

test/make_app_test.js
```diff
@@ -4,18 +4,18 @@ const makeApp = require('../lib/make_app').makeApp;
 const assert = require('chai').assert;
 const supertest = require('supertest');
 
-suite('makeApp', () => {
+suite.only('makeApp', () => {
```
rictic: remove `only`

src/make_app.ts
```diff
@@ -47,7 +50,9 @@ export function makeApp(options: AppOptions): PolyserveApplication {
 
   const app: PolyserveApplication = <PolyserveApplication>express();
 
-  app.get('*', function (req, res) {
+  app.use('*', babelCompile);
```
rictic: TODO: make this optional?

src/compile-middleware.ts
justinfagnani: done

src/compile-middleware.ts
justinfagnani: Is `-1` more clear than `null`? I like `null` to mean "missing value"

src/make_app.ts
justinfagnani: done

src/make_app.ts
justinfagnani: This issue is that if the browser sends a cache-control header because it has a cached version of the resource, and `send()` determines that the resource hasn't been modified, the response will be a 304 and won't contain a body or mime-type, and thus we can't compile it. This is actually the desired behavior for users, but during development it was massively hard to tell what was going on when I didn't know about this.

test/make_app_test.js
justinfagnani: done

justinfagnani:
I'll rebase in a sec

src/compile-middleware.ts
rictic: The tricky thing is that this requires the user to mentally model multi-type comparisons.

I had to spend a little while in a repl to mostly confirm to myself that `null` compares against numbers like it's `0` unless the comparison is `==` or `===`, but I'm still not entirely certain that that's always the case. The nice part about `-1` here is that it's very easy to tell how it will behave in the comparisons that follow.

justinfagnani: done

PR #117, opened 2016-11-19
Variants

rictic:


rictic:
Builds on top of #115

src/bower_config.ts
```diff
@@ -24,10 +24,9 @@ function bowerConfigContents(root?: string): string {
   let contents: string;
 
   try {
-    contents = fs.readFileSync(bowerConfigPath(root)).toString();
+    contents = fs.readFileSync(bowerConfigPath(root), 'utf-8');
```
FredKSchott: @TimvdLippe and I just merged a PR that makes a similar improvement, you may want to rebase/merge to pull in those changes

FredKSchott:
Could you update `polyserve.ts` to export the new public methods (`startServers()`)?

rictic:
Sure thing, done

src/start_server.ts
```diff
@@ -81,44 +81,153 @@ async function applyDefaultOptions(options: ServerOptions):
...
+    const app = getApp(options);
+    const server = await startWithPort(fullOptions, app);
+    return {app, server};
+  } catch (e) {
+    console.error('ERROR: Server failed to start:', e);
+    throw new Error(e);
+  }
+}
+
+export type ServerInfo = MainlineServer | VariantServer | DispatchServer;
```
FredKSchott: I'm currently left guessing about what role each of these servers performs purpose, if/how they interact, etc. Can you add some documentation to each type definition or elsewhere?

FredKSchott: 👍 perfect! thanks for the quick fixes

FredKSchott:
As discussed in person, added `headers` to `ServerOptions` to pass through to `makeApp()`/`getApp()`

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 * Compile JavaScript with Babel for browsers that don't support ES6
+* If the root dir contains directories that look like `bower_components-${foo}`, they will be treated as `variant dependencies`. In that case, polyserve will start one server for each variant, enabling testing and development of your code against each set of dependencies.
```
justinfagnani: WDYT of the phrase "dependency variants" and here, "treated as a `dependency variant`" ?

src/cli.ts
```diff
@@ -55,12 +60,32 @@ export async function run(): Promise<void> {
...
+  const serverInfos = await startServers(options);
+
+  if (serverInfos.length === 1) {
+    const urls = getServerUrls(options, serverInfos[0]!.server);
+    console.log(`Files in this directory are available under the following URLs
+    applications: ${
+                  url.format(urls.serverUrl)}
+    reusable components: ${url.format(urls.componentUrl)}
+  `)
+  } else {
```
justinfagnani: can you add a comment about what we're doing in the multiple servers case?

src/internal/next_open_port.ts
```diff
@@ -0,0 +1,34 @@
+/**
```
justinfagnani: I thought we moved internatl -> util?

src/cli.ts
```diff
@@ -55,12 +60,32 @@ export async function run(): Promise<void> {
...
+    const dispatchServer = serverInfos.find(s => s.kind === 'dispatch');
```
justinfagnani: Rather than return a simple array of servers, would it be nicer to return a struct that contains a reference to the dispatch server, app server and array of component servers? Are there any top-level properties / methods that would make sense?

src/cli.ts
```diff
@@ -55,12 +60,32 @@ export async function run(): Promise<void> {
...
+    if (!dispatchServer) {
+      throw new Error(
```
justinfagnani: Can this error be pushed down into `startServers`? Seems like this is saying part of it's contract is to start a dispatch server.

src/cli.ts
```diff
@@ -55,12 +60,32 @@ export async function run(): Promise<void> {
...
+          'Internal Error: Launched multiple servers but ' +
+          'didn\'t launch a dispatch server.');
+    }
+    const urls = getServerUrls(options, dispatchServer.server);
+    console.log(`Started multiple servers with different variants:
+    Dispatch server: ${url.format(urls.serverUrl)}`);
```
justinfagnani: Would a user know what a "dispatch server" is?

src/cli.ts
```diff
@@ -55,12 +60,32 @@ export async function run(): Promise<void> {
...
   }
-  else {
-    await startServer(options);
+
+  const serverInfos = await startServers(options);
+
+  if (serverInfos.length === 1) {
+    const urls = getServerUrls(options, serverInfos[0]!.server);
+    console.log(`Files in this directory are available under the following URLs
+    applications: ${
```
justinfagnani: why don't we always create a dispatch server?

src/make_app.ts
```diff
@@ -46,6 +46,8 @@ export interface PolyserveApplication extends express.Express {
  */
 export function makeApp(options: AppOptions): PolyserveApplication {
   options = options || {};
+  // Doing option fallback here and in start_server is awkward. We should have
```
justinfagnani: makes this a TODO

src/start_server.ts
```diff
@@ -81,44 +81,171 @@ async function applyDefaultOptions(options: ServerOptions):
         certPath: options.certPath || 'cert.pem',
         keyPath: options.keyPath || 'key.pem',
       });
+      withDefaults.packageName = options.packageName ||
```
justinfagnani: move into the `Object.assign` above

src/start_server.ts
```diff
@@ -81,44 +81,171 @@ async function applyDefaultOptions(options: ServerOptions):
...
+}
+
+export type ServerInfo = MainlineServer | VariantServer | DispatchServer;
+
+/**
+ * The `default` or `primary` server. If only one ServerInfo is returned from
+ * startServers it must be a MainlineServer. This is the server that's running
+ * with the default configuration and not running a variant configuration.
+ */
+export interface MainlineServer {
```
justinfagnani: Can we call this the `AppServer`?

src/start_server.ts
```diff
@@ -81,44 +81,171 @@ async function applyDefaultOptions(options: ServerOptions):
...
+ */
+export interface VariantServer {
+  kind: 'variant';
+  server: http.Server;
+  app: express.Application;
+  options: ServerOptions;
+  variantName: string;
+}
+/**
+ * If more than one server is started by startServers, the main port will serve
```
justinfagnani: I think we should always start this server, and "dispatch" might not be the best name. Maybe "control server" or something like that. I imaging this server growing in responsibility to do more than just dispatch to the other servers.

src/start_server.ts
```diff
@@ -81,44 +81,171 @@ async function applyDefaultOptions(options: ServerOptions):
...
+}
+
+/**
+ * Starts one or more web servers, based on the given options and
+ * variant bower_components directories that are found in the root dir.
+ */
+export async function startServers(options: ServerOptions):
+    Promise<ServerInfo[]> {
+  const root = options.root || process.cwd();
+  const filesInRoot = await fs.readdir(root);
```
justinfagnani: can you move this out into a `findVariants` function?

src/start_server.ts
```diff
@@ -81,44 +81,171 @@ async function applyDefaultOptions(options: ServerOptions):
...
+ * multiple variant dependency directories are detected/configured , there will
+ * be one MainlineServer that serves out the default dependency directory, and
+ * one VariantServer for each other dependency directory.
+ */
+export interface VariantServer {
+  kind: 'variant';
+  server: http.Server;
+  app: express.Application;
+  options: ServerOptions;
+  variantName: string;
```
justinfagnani: do we need `path` to point to which folder hosts the dependencies?

src/start_server.ts
```diff
@@ -81,44 +81,171 @@ async function applyDefaultOptions(options: ServerOptions):
...
+    kind: 'mainline',
+    server: mainServer.server,
+    app: mainServer.app,
+    options: mainlineOptions,
+  });
+
+  for (const variant of variantNames) {
+    const variantOpts = Object.assign({}, options);
+    variantOpts.port = 0;
+    variantOpts.componentDir = `bower_components-${variant}`;
```
justinfagnani: This hardcoding seems unfortunate. How about a description of a variant that included its path.

CHANGELOG.md
rictic: Sure, what do you think of this?

src/cli.ts
rictic: Because that seems disruptive. If you only need one server, why start a dispatch server to bounce you to that one server?

src/cli.ts
rictic: Changed to `More info here:`

src/internal/next_open_port.ts
rictic: Erp, yeah, that didn't make it down to this PR. Fixed and reformatted.

src/make_app.ts
rictic: Done

src/start_server.ts
rictic: I need `withDefaults.root`, so it needs to come after.

src/start_server.ts
rictic: It's not yet an AppServer is it? All of the VariantServers also serve out the application directory I think

src/start_server.ts
```diff
@@ -81,44 +81,171 @@ async function applyDefaultOptions(options: ServerOptions):
...
+ */
+export interface MainlineServer {
+  kind: 'mainline';
+  server: http.Server;
+  app: express.Application;
+  options: ServerOptions;
+}
+/**
+ * These are servers which are running some named variant configuration. For
+ * multiple variant dependency directories are detected/configured , there will
```
rictic: Done

src/start_server.ts
rictic: Renamed to `control` here and elsewhere, but I don't think we should disrupt our typical user uselessly. If we do add more features to the `control` server such that they'd be well served by this functionality, that seems like the time to start always starting a control server. If we do so now it's just annoying.

src/start_server.ts
rictic: Done

src/start_server.ts
rictic: +1, done

src/cli.ts
rictic: I was skeptical, but yes, much nicer.

src/cli.ts
rictic: Done

src/cli.ts
rictic: +1, pushed into the type system

src/start_server.ts
rictic: Added `dependencyDir`

src/start_server.ts
justinfagnani: If it's not the AppServer, what is it? Maybe that needs to be more clear.

rictic: It's the server that serves out the default dependencies. The one that's not a variant

src/start_server.ts
```diff
@@ -81,44 +80,193 @@ async function applyDefaultOptions(options: ServerOptions):
...
+}
+
+export interface MultipleServersInfo {
+  kind: 'MultipleServers';
+  mainline: MainlineServer;
+  variants: VariantServer[];
+  control: ControlServer;
+}
+
+export type StartServerResult = MainlineServer | MultipleServersInfo;
```
justinfagnani: To me is seems simpler to just have one return type, and that type can contain information on the different types of servers that were started - ie: rename `MultipleServersInfo` to `DevServers` or something and always return that.

src/start_server.ts
```diff
@@ -81,44 +80,193 @@ async function applyDefaultOptions(options: ServerOptions):
...
+export interface VariantServer {
+  kind: 'variant';
+  server: http.Server;
+  app: express.Application;
+  options: ServerOptions;
+  variantName: string;
+  dependencyDir: string;
+}
+/**
+ * If more than one server is started by startServers, the main port will serve
```
justinfagnani: I still this this particular variation in behavior based on the filesystem layout is itself potentially confusing, rather than always launching a control server and usually either an app server and/or components server. There are enough good things to add to a control server that I'm ok with claiming the space for it, with the added benefit on a single behavior to implement.

src/start_server.ts
```diff
@@ -81,44 +80,193 @@ async function applyDefaultOptions(options: ServerOptions):
...
+  app: express.Application;
+  options: ServerOptions;
+}
+/**
+ * These are servers which are running some named variant configuration. For
+ * multiple variant dependency directories are detected/configured, there will
+ * be one MainlineServer that serves out the default dependency directory, and
+ * one VariantServer for each other dependency directory.
+ */
+export interface VariantServer {
```
justinfagnani: Maybe name this ComponentServer. That it is for a particular variant is more of a side-effect of its `dependencyDir` property.

src/start_server.ts
justinfagnani: If it's not an app server, what is it? Maybe it needs a better name or more docs. I assume this is the server serving `/` and containing the 404-to-shell redirect behavior. The ComponentServers serve `/components-foo` and contain the `../` -> `/components` redirect behavior. Looking at rest of PR again to verify this.

src/test/cli_test.ts
```diff
@@ -14,19 +14,41 @@
...
+  // output.
+  // TODO(rictic): look into using spawn() to run polyserve in another process
+  //     instead. That way we could actually get it started and run it for a
+  //     while, hit it with requests, etc.
+  async function runCli(args: string[]) {
     const originalArgv = process.argv;
-    process.argv = ['node', 'polyserve', '--unknown-parameter'];
+    process.argv = ['node', 'polyserve'].concat(args);
+    let stdout = '';
+    let unintercept = intercept((txt) => {
```
justinfagnani: comment? what's this doing?

src/start_server.ts
rictic: Chatted about it over slack. My understanding is that we agreed that we'll likely want to do this later, but as of now the control server isn't doing enough to be worth putting in front of users without variants. We can always reverse this later as the control server does more.

src/start_server.ts
```diff
@@ -81,44 +80,193 @@ async function applyDefaultOptions(options: ServerOptions):
...
+  //   metadata about them besides their names?
+  if (variants.length > 0) {
+    return await startVariants(options, variants);
+  }
+
+  const serverAndApp = await _startServer(options);
+  return {
+    kind: 'mainline',
+    server: serverAndApp.server,
+    app: serverAndApp.app, options
```
justinfagnani: nit: I thought our style was to add trailing commas. Also, `options` not being on its own line makes it really easy to miss.

rictic: Ah, this repo doesn't have a tslint. I'll add that in a follow up PR.

Moved options up so that it's on its own line

PR #120, opened 2016-11-22
add ability to serve user-provided routes before components/ handler

FredKSchott:
This is a fork of @rictic's variant's branch which adds the ability for the user to define their own routes. This solves the problem where WCT needs to serve the generated webrunner out of the `components` directory, but polyserve handles all `components/` routes and returns a 404 because the file doesn't actually exist on disk. Because the response is handled, any later-attached route handlers never fire.

This was all we need to get wct running again. But I'd rather see this solved by having polyserve call next() inside it's route handler if the file was not found. This would give consumers the ability to attach routes to the `components/` directory to handle routes that polyserve can't. This would also solve the "generated test runner" problem.

/cc @rictic @justinfagnani

### DIFF
--- a/src/make_app.ts
+++ b/src/make_app.ts
@@ -48,7 +48,10 @@ export function makeApp(options: AppOptions): PolyserveApplication {
   // TODO(rictic): Doing option fallback here and in start_server is awkward. We
   // should have just one.
   const root = options.root;
-  const componentDir = options.componentDir || 'bower_components';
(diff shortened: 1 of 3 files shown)


rictic:
Rebased against master. PTAL @justinfagnani

rictic:
Squashed down the intermediate commits. Will merge once travis is green.

PR #122, opened 2016-11-23
Add tslint

rictic:
- [x] CHANGELOG.md not updated, internal change

### DIFF
--- a/src/compile-middleware.ts
+++ b/src/compile-middleware.ts
@@ -57,7 +57,7 @@ export function babelCompile(forceCompile: boolean): RequestHandler {
           compileMimeTypes.indexOf(getContentType(response)) >= 0;
     },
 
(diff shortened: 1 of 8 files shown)


src/compile-middleware.ts
```diff
@@ -52,7 +52,7 @@ export const babelCompile: RequestHandler = transformResponse({
         compileMimeTypes.indexOf(getContentType(response)) >= 0;
   },
 
-  transform(request: Request, response: Response, body: string): string{
+  transform(request: Request, response: Response, body: string): string /* */ {
```
justinfagnani: what's this?

rictic: clang doing something weird with method literal syntax in an object literal

rictic: This is still going on, I just had to modify it slightly for clang to not wrap the line.

PR #124, opened 2016-11-23
Add --compile flag

justinfagnani:
Adds a `--compile` flag with valid values of 'always', 'never', and 'auto', and a default of 'auto'.


justinfagnani:
Changelog update incoming

rictic:
Please rebase against master as well

src/test/make_app_test.ts
```diff
@@ -14,6 +14,7 @@
```
rictic: Please delete `/test/make_app_test.js`

src/test/make_app_test.ts
```diff
@@ -47,12 +48,82 @@ suite('makeApp', () => {
...
+  suite('compilation', () => {
+
+    const testCompilation = ({url, agent, compile, result}: {
+      url: string,
+      agent?: string,
+      compile: 'always' | 'never' | 'auto',
+      result: 'compiled' | 'uncompiled'
+    }) => async() => {
+      const app = makeApp({
+        root: __dirname,
```
rictic: root should probably be `root` here rather than `__dirname`

src/test/make_app_test.ts
```diff
@@ -47,12 +48,82 @@ suite('makeApp', () => {
...
+        componentDir: path.join(root, 'bower_components'), compile,
+      });
+      let tester = supertest(app).get(url);
+      if (agent) {
+        tester = tester.set('User-Agent', agent)
+      }
+      await tester.expect((res: Response) => {
```
rictic: can be rewritten more clearly as:

```typescript
const res = await tester;
// contents of expect callback here
```

src/compile-middleware.ts
```diff
@@ -42,66 +46,71 @@ function isSuccessful(response: Response) {
...
+          compileMimeTypes.indexOf(getContentType(response)) >= 0;
+    },
+
+    transform(request: Request, response: Response, body: string): string{
+      const contentType = getContentType(response);
+      const uaParser = new UAParser(request.headers['user-agent']);
+      const compile = forceCompile || needCompilation(uaParser);
+
+      if (compile) {
+        // TODO(justinfagnani): cache compilation results
```
rictic: For development it may be that the built-in HTTP caching that we're getting with send() should make this just fine. It won't help at all with multiple clients, but it'll work fine for a single client.

src/compile-middleware.ts
```diff
@@ -46,47 +46,51 @@ function isSuccessful(response: Response) {
...
-
-    if (compile) {
-      // TODO(justinfagnani): cache compilation results
-      if (contentType === htmlMimeType) {
-        const document = parse5.parse(body);
-        const scriptTags = dom5.queryAll(document, isInlineJavaScript);
-        for (const scriptTag of scriptTags) {
+export function babelCompile(forceCompile: boolean): RequestHandler {
+  if (forceCompile == null)
+    forceCompile = false;
```
rictic: brackets please

justinfagnani:
rebased, formatted and updated CHANGELOG

src/test/make_app_test.ts
```diff
@@ -47,12 +48,82 @@ suite('makeApp', () => {
...
       called = true;
     };
     const app = makeApp({root: path.resolve(__dirname, 'no_bower_json/')});
     assert.isFalse(called);
     assert.equal(app.packageName, 'no_bower_json');
   });
 
+  suite('compilation', () => {
+
+    const testCompilation = ({url, agent, compile, result}: {
```
rictic: Destructuring crashes node 4

src/test/make_app_test.ts
justinfagnani: that's... odd. `supertest.Test` is a `Test`, `Request`, _and_ a `Promise<Response>`? 🙄

justinfagnani: but ok

src/test/make_app_test.ts
justinfagnani: done

src/test/make_app_test.ts
justinfagnani: done

justinfagnani:
PTAL

src/compile-middleware.ts
rictic: We should see about getting tslint to enforce brackets around the then clause. Mind adding them here?

justinfagnani: done. I had written this as a single-line if, which I'm in favor of, so we should also see if clang can let those through.

PR #126, opened 2016-11-27
Start servers fix

### LINKED ISSUE #125: PR #117 broke polyserve on the CLI
Running `polyserve` with the CLI is broken since #117 has been merged.
I have bisected from the previous release to master and the merge commit of #117 has caused that running `./bin/polyserve` directly terminates after logging `no bower.json detected, using name "serve"` to the console. Polyserve is not running and can not process any requests.

justinfagnani:
Fixes #125

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -18,13 +18,13 @@ import * as path from 'path';
 import * as url from 'url';
 
 import {args} from './args';
-import {getServerUrls, ServerOptions, startServers} from './start_server';
+import {getServerUrls, ServerOptions, StartServerResult, startServers} from './start_server';
 
 import commandLineArgs = require('command-line-args');
 import commandLineUsage = require('command-line-usage');
 
 
-export async function run(): Promise<void> {
+export async function run(): Promise<StartServerResult> {
   const argsWithHelp: ArgDescriptor[] = args.concat({
     name: 'help',
     description: 'Shows this help message',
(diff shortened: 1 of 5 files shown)


src/cli.ts
```diff
@@ -67,7 +67,8 @@ export async function run(): Promise<void> {
   const serverInfos = await startServers(options);
 
   if (serverInfos.kind === 'mainline') {
-    const urls = getServerUrls(options, serverInfos[0]!.server);
+    const mainlineServer = serverInfos;
```
justinfagnani: This is the actual fix. The rests is testing.

src/test/cli_test.ts
```diff
@@ -40,8 +41,21 @@ suite('cli', () => {
...
 
+    const closeServer = (server: http.Server) =>
+        new Promise((resolve, reject) => server.close((e: Error) => {
+          e ? reject(e) : resolve();
+        }));
+
     try {
-      await cliRun();
+      const serverInfos = await cliRun();
+      if (serverInfos) {
```
justinfagnani: Now that there's a test that tries to successfully launch some servers, we need to shut them down and wait or else we get port-in-use errors.

src/cli.ts
rictic: Ouch. I guess we get that from not having strict indexing types. You don't need to do this assignment, though it does help legibility.

TimvdLippe:
Described a different issue regarding `filePath` resolution with `send()` in #125 too, but at least this PR fixes invoking the CLI.

PR #127, opened 2016-11-28
Change default bind address from localhost to 0.0.0.0

gtirloni:
- [X] CHANGELOG.md has been updated

This is to work around an issue running `polymer serve` inside Docker containers, as explained in polymer/polymer-cli#300. It follows the same behavior as node's [http.Server.listen](https://nodejs.org/api/http.html#http_server_listen_port_hostname_backlog_callback).

With this change in place, `polymer serve` works inside a Docker container without having to use `-H 0.0.0.0`. 

One security drawback is that, in case a developer doesn't have a firewall in his/her laptop, this would expose the service. In my team's environment that would be okay (everybody has a firewall that defaults to block), but I'm not sure about others.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 `--compile` flag controls compilation, with valid values of 'always', 'never',
 and 'auto'; and a default of 'auto'.
 * If the root dir contains directories that look like `bower_components-${foo}`, they will be treated as `dependency variants`. In that case, polyserve will start one server for each variant, enabling testing and development of your code against each set of dependencies.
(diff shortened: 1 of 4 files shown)


justinfagnani:
I don't think we want to make this change. As a local development server, `localhost` is much more secure against accidentally serving up filesystem contents to the local network, or possibly the whole internet.

I think that users who want to test on a non-local device, or those who may want to try polyserve as a production server (which isn't really a primary use case), can opt-in to a specific `host` setting and hopefully understand the security implications.

gtirloni:
Agreed. Is the `host` setting something that can be set on the polymer.json file or only through `-H 0.0.0.0`?

justinfagnani:
Not through polymer.json yet, but that may be a good middle-ground here.

PR #129, opened 2016-11-28
Compile fixes

justinfagnani:


justinfagnani:
No tests for these, but we need this fix quick for elements.

justinfagnani:
Merging TBR

src/compile-middleware.ts
```diff
@@ -21,7 +21,27 @@ import {UAParser} from 'ua-parser-js';
...
+  'babel-plugin-transform-es2015-literals',
+  'babel-plugin-transform-es2015-object-super',
+  'babel-plugin-transform-es2015-parameters',
+  'babel-plugin-transform-es2015-shorthand-properties',
+  'babel-plugin-transform-es2015-spread',
+  'babel-plugin-transform-es2015-sticky-regex',
+  'babel-plugin-transform-es2015-template-literals',
+  'babel-plugin-transform-es2015-typeof-symbol',
+  'babel-plugin-transform-es2015-unicode-regex',
+  'babel-plugin-transform-regenerator',
```
rictic: Why this list rather than babelLatest?

justinfagnani: babel-latest includes a module transform, which assumes that input is an ES6 module and compiles top-level `this` to `undefined` :(

rictic: yikes!

PR #130, opened 2016-12-02
Update args to be compatible with polymer-cli

justinfagnani:

### DIFF
--- a/src/args.ts
+++ b/src/args.ts
@@ -12,20 +12,18 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-/// <reference path="../custom_typings/command-line-args.d.ts" />
-
 import {ArgDescriptor} from 'command-line-args';
 
 export let args: ArgDescriptor[] = [
   {
     name: 'version',
-    alias: 'v',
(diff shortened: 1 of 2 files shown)


src/args.ts
```diff
@@ -12,20 +12,18 @@
...
   {
     name: 'version',
-    alias: 'v',
     description: 'Print version info',
     type: Boolean,
   },
   {
     name: 'root',
-    description: 'The root directory. Defaults to cwd',
+    description: 'The directory where your project exists. Defaults to the ' +
```
rictic: `The directory where your project exists` is kinda roundabout and not super clear. What about:

```javascript
'The root directory of your project. Defaults to the current working directory'
```

justinfagnani: This has to match what's in the CLI, or the validation fails.

justinfagnani: I'd like to come up with a better way to share argument definitions across repos, but not today :)

justinfagnani: Actually, since I'm about to push a PR on CLI, I can change this now.

PR #131, opened 2016-12-02
Fix proxy tests

justinfagnani:
The probably was port contention between the proxied test server and polyserve. I also made teardown wait for servers to close.

### DIFF
--- a/src/test/start_server_test.ts
+++ b/src/test/start_server_test.ts
@@ -84,6 +84,7 @@ suite('startServer', () => {
       app = await startServer({root});
 
       return proxyServer = await startServer({
+          port: 0,
           root: __dirname,
           proxy: {
             path: path,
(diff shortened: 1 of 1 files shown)


src/test/start_server_test.ts
```diff
@@ -97,11 +98,20 @@ suite('startServer', () => {
...
+      function _done() {
+        if (--waitingCount === 0) {
+          done();
+        }
+      }
+      if (proxyServer) {
+        waitingCount++;
+        proxyServer.close(_done);
+      }
+      app.close(_done);
```
rictic: I think we should strongly prefer async/await over callback counting for all the usual reasons.

justinfagnani: to use async await I have to convert to promises first

TimvdLippe:
I was not aware of any failures, neither locally nor on Travis. Did I miss something? Thanks for fixing anyways!

rictic:
Yeah, it's kinda strange. It never failed on travis, but both Justin and I were seeing failures locally.

¯\_(ツ)_/¯

PR #133, opened 2016-12-03
Add another custom_typings ref for definitions generation.

rictic:


justinfagnani:
I removed this reference because of a conflict with polymer-cli. Does the CLI still build without errors with this change?

rictic:
I `npm link`ed this into polymer-cli master and `tsc` didn't report any conflicts (it did notice a null violation in `build/build.ts` but that seems unrelated.

That's not a 100% guarantee that it won't conflict after release though, as `npm link` isn't the same as a normal install, but this is more correct I think, and if there are issues in polymer-cli after release I'll fix those there.

PR #135, opened 2016-12-03
Implement websocket connection to implement live reload

### LINKED ISSUE #26: livereload
Could this tool have optional livereload?:
https://github.com/hiddentao/gulp-server-livereload

### LINKED ISSUE #86: Add BrowserSync?
I use Polyserve while developing quite a lot, using the demo page as a way of testing output. I don't know how wide-spread a use case that is, but it would be really nice if Polyserve had BrowserSync added so as to gain the benefit of multi-device events and automatic refresh. 

Happy to try doing a PR if this would be useful.

TimvdLippe:
This PR implements live-reload with Websockets. It uses `chokidar` to watch for files and `ws` to construct the server-side websocket. A request to `host:port/live-reload.js` will install a simple websocket which listens to any message of the server. The server will send a message when a file was changed in the `live-reload-path`.

You can test this branch with `./bin/polyserve static/ --live-reload-path src/`. Open a browser on the one side of your window and a text editor on the other side. Load `localhost:8080` in the browser and modify a file in the `src` folder of `polyserve`. You can see that the page has reloaded.

Before I did too much work, I wanted to know if you approve this implementation. If you do, I can easily add tests and possibly additional work.

One thing I was wondering about: could we inject the client-side js (essentially a one-liner) by modifying `index.html` with `dom5`. Just like the files are transformed with `babel` compilation. This would remove the special `/live-reload.js` handler, but probably requires some work to make the transformation work.

 - [x] Write tests

Fixes #26
Fixes #86 
Fixes https://github.com/Polymer/polymer-cli/issues/230

---
This change is [Reviewable](https://reviewable.io/reviews/polymerlabs/polyserve/135)

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "@types/express": "^4.0.33",
     "@types/mime": "0.0.29",
     "@types/mz": "0.0.29",
-    "@types/node": "^4.0.30",
+    "@types/node": "=4.0.30",
     "@types/opn": "^3.0.28",
     "@types/parse5": "^2.2.34",
     "@types/pem": "^1.8.1",
@@ -54,6 +54,7 @@
     "babel-plugin-transform-es2015-typeof-symbol": "^6.18.0",
     "babel-plugin-transform-es2015-unicode-regex": "^6.11.0",
     "babel-plugin-transform-regenerator": "^6.16.1",
+    "chokidar": "^1.6.1",
     "command-line-args": "^3.0.1",
     "command-line-usage": "^3.0.3",
     "content-type": "^1.0.2",
@@ -70,13 +71,15 @@
     "resolve": "^1.0.0",
     "send": "^0.14.1",
     "spdy": "^3.3.3",
-    "ua-parser-js": "^0.7.12"
+    "ua-parser-js": "^0.7.12",
+    "ws": "^1.1.1"
   },
   "devDependencies": {
     "@types/assert": "0.0.29",
     "@types/chai": "^3.4.34",
     "@types/chai-as-promised": "0.0.29",
     "@types/lru-cache": "^2.5.32",
+    "@types/chokidar": "^1.4.31",
     "@types/mocha": "^2.2.33",
     "@types/sinon": "^1.16.31",
     "@types/supertest": "^1.1.31",
--- a/src/start_server.ts
+++ b/src/start_server.ts
@@ -13,6 +13,7 @@
  */
 
 import * as assert from 'assert';
+import {watch} from 'chokidar';
 import * as express from 'express';
 import * as fs from 'mz/fs';
 import * as path from 'path';
@@ -21,6 +22,7 @@ import * as send from 'send';
 // https://github.com/molnarg/node-http2/issues/100
 import * as http from 'spdy';
 import * as url from 'url';
+import {Server as WebSocketServer} from 'ws';
 
 import {bowerConfig} from './bower_config';
 import {babelCompile} from './compile-middleware';
@@ -83,6 +85,10 @@ export interface ServerOptions {
   /** An optional list of routes & route handlers to attach to the polyserve
    * app, to be handled before all others */
   additionalRoutes?: Map<string, express.RequestHandler>;
+
+  /** An optional file path to watch for to automatically send websocket
+   * messages to notify connected browsers */
+  liveReloadPath?: string;
 }
 
 function applyDefaultServerOptions(options: ServerOptions) {
@@ -350,6 +356,12 @@ export function getApp(options: ServerOptions): express.Express {
     app.use('*', babelCompile(options.compile === 'always'));
   }
 
+  if (options.liveReloadPath) {
+    app.get('/_polyserve/live-reload.html', (_, res) => {
+      res.sendFile(path.resolve(__dirname, '..', 'static', 'auto-reload.html'));
+    });
+  }
+
   app.use(`/${componentUrl}/`, polyserve);
 
   app.get('/*', (req, res) => {
@@ -478,6 +490,21 @@ async function startWithFirstAvailablePort(
 async function tryStartWithPort(
     options: ServerOptions, app: express.Application, port: number) {
   const server = await createServer(app, options);
+
+  if (options.liveReloadPath) {
+    const wss = new WebSocketServer({server: server, clientTracking: true});
+    const watcher = watch(options.liveReloadPath);
+    watcher.on('change', () => {
+      wss.clients.forEach((ws) => {
+        ws.send('changed');
+      });
+    });
+    server.on('close', () => {
+      wss.close();
+      watcher.close();
+    });
+  }
+
   return new Promise<http.Server|null>((resolve, _reject) => {
     server.listen(port, options.hostname, () => {
       resolve(server);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 * Added `-u`/`--component-url` option to support expressing different url to
   fetch components from than `components`.
+* Added `--live-reload-path` option to automatically reload webpages during development.
 
 ### Fixed
 * When no port is given, do a better job of finding an available port.
--- a/custom_typings/ws.d.ts
+++ b/custom_typings/ws.d.ts
@@ -1,0 +1,23 @@
+declare module 'ws' {
+  import * as http from 'spdy';
+  interface WebSocketServerOptions {
+    server: http.Server;
+    clientTracking: boolean;
+  }
+  export class Server {
+    constructor(options: WebSocketServerOptions)
+    clients: WebSocket[];
+    on(event: 'connection', cb: (client: WebSocket) => void): this;
+    close(cb?: () => void): this;
+  }
+  export class WebSocket {
+    constructor(address: string)
+    status: number;
+    message: string;
+    send(data: any, cb?: (err: Error) => void): void;
+    on(event: 'message', cb: (data: any, flags: {}) => any): void;
+    on(event: 'close', cb: () => any): void;
+  }
+
+  export default WebSocket;
+}
--- a/src/args.ts
+++ b/src/args.ts
@@ -139,4 +139,9 @@ export let args: ArgDescriptor[] = [
         'Host URL to proxy to, for example `https://myredirect:8080/foo`',
     type: String,
   },
+  {
+    name: 'live-reload-path',
+    description: 'The file path to watch for changes to automatically reload connected browsers.',
+    type: String,
+  },
 ];
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -64,6 +64,7 @@ export async function run(): Promise<StartServerResult> {
     certPath: cliOptions['cert'],
     pushManifestPath: cliOptions['manifest'],
     proxy: proxyArgs.path && proxyArgs.target && proxyArgs,
+    liveReloadPath: cliOptions['live-reload-path'],
   };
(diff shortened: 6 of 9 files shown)


src/start_server.ts
```diff
@@ -456,6 +468,23 @@ async function startWithFirstAvailablePort(
 async function tryStartWithPort(
     options: ServerOptions, app: express.Application, port: number) {
   const server = await createServer(app, options);
+
+  if (options.liveReloadPath) {
+    const wss = new WebSocketServer({server: server});
+    wss.on('connection', (ws: WebSocket) => {
+      const watcher = watchFile(options.liveReloadPath);
```
rictic: If `liveReloadPath` points to a directory it would be nice to recursively watch that directory and its children. I suspect that watchFile when given a directory will see changes to that directory (in some filesystems at least) but not changes inside subdirectories.

rictic: As written I think this creates another file watcher every time the browser reloads. Depending on how watchFile is implemented it may be much more efficient to create the watcher once and just add and remove event listeners as sockets open and close.

justinfagnani:
Go away googlebot.

justinfagnani:
@TimvdLippe this is very interesting. My big question here, since I don't know much about live reload solutions in node, is: is this the idiomatic or currently best way to do this? What are the other options? Why not BrowserSync?

If you could help answer those it would let us move forward. Thanks!

rictic (commented):
This seems reasonable to me. `live-reload.js` seems minimal enough that it doesn't seem crazy to inject into pages, the tricky thing would be either injecting it only into the topmost page (and not html files that have been imported). That's probably not feasible. Hm. Instead we could write the client side script such that if it has already run in the page from another file it's a noop. Something like:

```javascript
if (!window.__liveReloadSocket && window.WebSocket) {
  window.__liveReloadSocket = new WebSocket("ws://" + window.location.hostname + ":" + window.location.port);
  window.__liveReloadSocket.onmessage = function() {
    window.location.reload();
  };
}
```

Or, perhaps better still, make it `live-reload.html` and inject html imports onto the end of files. Might not even have to do any clever `dom5` stuff, but just like `response.send(responseText + '\n<link rel="import" href="/_polyserve/live-reload.html">');`

src/start_server.ts
```diff
@@ -347,11 +354,19 @@ export function getApp(options: ServerOptions): express.Express {
   }
 
   if (options.compile === 'auto' || options.compile === 'always') {
-    app.use('*', babelCompile(options.compile === 'always'));
+    // Compile all files, except index.html
+    app.use(/\/.+/, babelCompile(options.compile === 'always'));
```
TimvdLippe: This was required to not break serving `index.html`. When using 2 transformers, calling `res.end()` twice will render an invalid response and the browser shows a white page.

src/live-reload-middleware.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import {transformResponse, getContentType, isSuccessful} from './transform-middleware';
+
+export const liveReloadAppend: RequestHandler =
+  transformResponse({
+    shouldTransform(_request: Request, response: Response) {
+      return isSuccessful(response) && getContentType(response) === 'text/html';
+    },
+
+    transform(_request: Request, _response: Response, body: string): string {
```
TimvdLippe: This transformation is required to not generate invalid HTML. Appending the import simply to body resulted in invalid HTML which the browser did not want to parse correctly.

TimvdLippe:
@justinfagnani Neither do I, the WebSocket solution was originally my own. However, I took a look at how `BrowserSync` implements live-reload and they appear to be using sockets too: https://github.com/BrowserSync/browser-sync/blob/master/lib/sockets.js

Including `BrowserSync` into `polyserve` seems unnecessary, as most of the features of `BrowserSync` are already implemented in `polyserve`. I chose to implement it directly in `polyserve` since it is a very easy and small change. Importing a library that has too many/duplicate features for `polyserve` is wasteful imo.

src/start_server.ts
TimvdLippe: The naming was confusing as I aliased `watch` to `watchFile` which is actually incorrect. I changed this back to just `watch`, as passing directories/files/globs etc... will all be handled by `chokidar`. E.g. when passing a directory such as `src/`, chokidar will listen for sub-directories.

TimvdLippe:
Took a shower and thought: what if we pass an array of transformers to the middleware? That would solve quite some issues, because I also thought that it does not work with an index.html in a subdirectory (e.g. demo/index.html)

package.json
```diff
@@ -70,13 +71,15 @@
...
-    "ua-parser-js": "^0.7.12"
+    "ua-parser-js": "^0.7.12",
+    "ws": "^1.1.1"
   },
   "devDependencies": {
     "@types/assert": "0.0.29",
     "@types/chai": "^3.4.34",
     "@types/chai-as-promised": "0.0.29",
     "@types/lru-cache": "^2.5.32",
+    "@types/chokidar": "=1.4.29",
```
TimvdLippe: For more discussion on this specific version see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/13126#discussion_r92944133

TimvdLippe:
I decided to give it a try with an array of transformers passed to the `transform-middleware`. Therefore `text/html` responses can be altered with `dom5` by passing the dom to each transformer. This prevents needless serialization. This also fixes the issue where the `live-reload-transformer` could only transform `index.html` and not nested `index.html`.

The only thing I am not 100% happy with is the serialization/parsing required to store/retrieve HTML documents from the `babelCache`. A solution could be to have 2 separate caches, but not sure what your standpoint is on this.

Looking forward to your response!

FredKSchott:
This is ancient history, but I made a PR to add browser-sync way back when: https://github.com/PolymerLabs/polyserve/commit/9249cd882587eb3341c463300374ab101c347ab7. (We decided to hold off because we weren't sure whether or not polyserve was going to be replaced).

If it still works with the new variants system, it would be a lot less overhead to use browser-sync than writing and maintaining our own websocket connection. Web sockets are a yuuuuuge pain to debug ime.

TimvdLippe:
@FredKSchott The curent implementation is just ~10 lines of code: https://github.com/PolymerLabs/polyserve/pull/135/files#diff-7a0887e23c118207eae676788b29e7e8R503

The reason I would vouch for this implementation is that it is a very simple change. Adopting a dependency for the 1 function it provides, namely reloading, seems counter-intuitive for me. Browsersync is a fairly large package which does not only offer live reloading, but serving, proxy, etc... Instead of maintaining a websocket implementation, the dependency has to be maintained.

This PR contains a lot of other modifications to make the automatic injection of the HTML import work. It is not as easy as I expected it to be so I can remove these parts from the PR if you want to. The size of the PR will then be comparable to the commit you linked.

FredKSchott:
Gave this a closer look, I'd still prefer to us delegate our live-reload implementation to an external library rather than create and maintain our own. While you're right that the `if (options.liveReloadPath)` conditional is small, I disagree that it is a simple change. You've also had to add new logic to watch for changes, manage transformers, `live-reload-middleware.js`, `transform-middleware.js`, etc.

Either way, I'll leave this to @rictic or @justinfagnani who have spent the most time in this repo recently. I've voiced my concerns here and would respect either of their decisions on this.

FredKSchott (commented):
Gave this a closer look, I'd still prefer to us delegate our live-reload implementation to an external library rather than create and maintain our own. While you're right that the if (options.liveReloadPath) conditional is small, I disagree that it is a simple change. You've also had to add new logic to watch for changes, manage transformers, live-reload-middleware.js, transform-middleware.js, etc.

Either way, I'll leave this to @rictic or @justinfagnani who have spent the most time in this repo recently. I've voiced my concerns here and would respect either of their decisions on this.

TimvdLippe:
@FredKSchott Yes agreed that the logic to "automagically inject the websocket script" is non-trivial. I will remove that part from this PR to reduce the amount of complexity and then you can reevaluate whether you still want to keep it or close it. I completely understand if in the end live-reload is implemented using an external library.

TimvdLippe:
Okay I removed a lot of the cruft. The implementation should now be a lot less intrusive :smile:

package.json
```diff
@@ -27,7 +27,7 @@
     "@types/express": "^4.0.33",
     "@types/mime": "0.0.29",
     "@types/mz": "0.0.29",
-    "@types/node": "^4.0.30",
+    "@types/node": "=4.0.30",
```
TimvdLippe: Installing locally throws various typing errors with `error TS2304: Cannot find name 'console'.` in a large number of files. It seems that version `4.2.0` of the typings are breaking installation. Therefore I pinned this version to the last version before `4.2.0`

tony19: I just tried this change, but it causes a [test failure in Node 4.7.2](https://travis-ci.org/Polymer/polymer-build/jobs/191794387#L191) (at least for the `polymer-build` repo):

```shell
[23:08:04] SyntaxError in plugin 'gulp-mocha'
Message:
    Unexpected token ...
Stack:
SyntaxError: Unexpected token ...
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:373:25)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at /home/travis/build/Polymer/polymer-build/node_modules/gulp-mocha/node_modules/mocha/lib/mocha.js:220:27
    at Array.forEach (native)
    at Mocha.loadFiles (/home/travis/build/Polymer/polymer-build/node_modules/gulp-mocha/node_modules/mocha/lib/mocha.js:217:14)
```

TimvdLippe: It seems that this package is breaking quite some.other project builds too. Had the same issue with my dom5 PR

tony19: Turns out my error had nothing to do with this change :)

It was actually caused by Node 4 not supporting the ES6 spread operator.

martijnjanssen: Isn't it true that the `=` is not a valid option in the depedencies? To have an exact version, there should only be a version. I couldn't find it in the [docs](https://docs.npmjs.com/files/package.json#dependencies).

TimvdLippe: Hm yes correct, for some reason I found this solution in a different package that was affected by this regression. I did not check the docs. Anyways, #155 solves the issue so I will need to rebase after that.

camerow:
Let's get this merged!

abdonrd:
Any news about this?

yveslange:
Hi, can you resolve the small conflict so this can be merged ? <3

TimvdLippe:
@yveslange I dont think this one will be merged, given the lack of response. Keeping it open to signal I still want to see it merged. If merging becomes likely, I will resolve potential merge conflicts.

shinxi:
Why not get it merged... Live reload is a cool feature! That'll be good news for polymer developers.

TimvdLippe:
@justinfagnani @rictic Just talked with @FredKSchott. His last comment still stands, so I would like to hear your opinion. If you agree that this is not the solution you are looking for, then let's close this PR 😄

abdonrd:
@justinfagnani @rictic @FredKSchott Any news? 😃

justinfagnani:
@abdonrd I'd like to talk to @TimvdLippe about this a little bit. Tim, maybe we can chat this week?

TimvdLippe:
@justinfagnani Um sure. Actually this evening (your morning) is pretty much the last evening I am avilable this week. Would that work for you?

TimvdLippe:
This PR is very old and I am no longer pursuing this change. Therefore I am closing this PR.

PR #136, opened 2016-12-05
Extend command-line-args typing for cli

rictic:
With this we can delete the custom_typings from cli and there's no conflict. Will upstream all of this the moment the testing fire is out.

### DIFF
--- a/custom_typings/command-line-args.d.ts
+++ b/custom_typings/command-line-args.d.ts
@@ -12,8 +12,14 @@ declare module 'command-line-args' {
     }
   }
 
+  /**
+   * @param descriptors An array of objects that describe the arguments that
+   *     we want to parse.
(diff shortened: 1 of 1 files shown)


custom_typings/command-line-args.d.ts
```diff
@@ -10,8 +8,12 @@ declare module 'command-line-args' {
       type?: Object;
       multiple?: boolean;
       defaultOption?: boolean;
+      group?: string;
     }
   }
 
+  function commandLineArgs(
+      groups: commandLineArgs.ArgDescriptor[], args?: string[]): any;
```
justinfagnani: docs? what's the last arg?

rictic: Documented.

PR #142, opened 2016-12-09
Added -u/--component-url feature.  (Also deduplicated logic for applying default server options.)

usergenic:
 - Added the `-u`/`--component-url` flag to allow you to specify something other than `components` as the URL to serve the components out of.  A partner option for `--component-dir` but they don't currently inform each other.

I think if we did a major release, polyserve should honor the `directory` flag of the root's `.bowerrc` and in the absense of that default `--component-dir` and `--component-url` to bower's default of `bower_components`  -- *at least* when a bower.json file is present, indicating polyserve is serving an app that uses bower.

### DIFF
--- a/src/start_server.ts
+++ b/src/start_server.ts
@@ -58,6 +58,9 @@ export interface ServerOptions {
   /** The component directory to use **/
   componentDir?: string;
 
+  /** The component url to serve **/
+  componentUrl?: string;
+
   /** The package name to use for the root directory **/
   packageName?: string;
 
@@ -81,7 +84,7 @@ export interface ServerOptions {
   additionalRoutes?: Map<string, express.RequestHandler>;
 }
 
-function applyDefaultOptions(options: ServerOptions) {
+function applyDefaultServerOptions(options: ServerOptions) {
   const withDefaults: ServerOptions = Object.assign({}, options);
   Object.assign(withDefaults, {
     port: options.port || 0,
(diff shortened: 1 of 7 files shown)


justinfagnani:
So I've been wanting to separate the app serving and component serving functionality so each server has a limited amount of magic going on. Polyserve's component mapping shouldn't be necessary in the application case since `../` URLs won't be used, and it adds a new route that isn't really valid, which is `/components/{app-name}/` -> `/`. That won't exist on a regular server.

usergenic:
Nearly all apps will have a components folder however.  Usually its called 'bower_components' which is the reason for supporting `-u`.  Polyserve already makes it possible to do app development while simultaneously building or testing the components it depends on by way of `--components-dir` to some other place.  It just doesn't conveniently allow you to serve the components via the name of the component directory you're actually using.

usergenic:
I think if you're saying you *don't* want to support application serving with polyserve, that also seems counter-intuitive, but I'm happy to wait until we can have a fuller conversation on it.  I may have mistaken this for an obvious net-positive feature, but I've not been part of the polyserve story and I know there's been a lot of thinking around its features esp. wrt variants etc, so this might run counter to it in ways I don't see.

justinfagnani:
My ideas around separating the magic of app and component serving are vague enough that it's not worth blocking on...

src/start_server.ts
```diff
@@ -81,14 +84,20 @@ export interface ServerOptions {
...
   const withDefaults: ServerOptions = Object.assign({}, options);
   Object.assign(withDefaults, {
     port: options.port || 0,
     hostname: options.hostname || 'localhost',
     root: path.resolve(options.root || '.'),
     certPath: options.certPath || 'cert.pem',
     keyPath: options.keyPath || 'key.pem',
+    // TODO(usergenic): The current behavior of polyserve is to use
+    // bower_components for directory and components for url. We should
+    // honor the value of `directory` of `.bowerrc` if found in the root dir
```
justinfagnani: I would make an exception for `directory` values that reach out of the project directory. That should require a flag given how dangerous it is.

usergenic:
Thank you for approving this PR!  :)

PR #144, opened 2016-12-10
Added an LRU cache for babel compilation.  With tests!

usergenic:
When polyserve compiles, it uses a cache to store the compiled output, hopefully making things faster for people.


src/compile-middleware.ts
```diff
@@ -16,11 +16,13 @@ import * as babelCore from 'babel-core';
 import {parse as parseContentType} from 'content-type';
 import * as dom5 from 'dom5';
 import {Request, RequestHandler, Response} from 'express';
+import * as LRU from 'lru-cache';
 import * as parse5 from 'parse5';
 import {UAParser} from 'ua-parser-js';
 
 import {transformResponse} from './transform-middleware';
 
+
```
usergenic: This was an unncessary newline add.  I'll remove.

src/compile-middleware.ts
```diff
@@ -66,6 +68,11 @@ function isSuccessful(response: Response) {
   return (statusCode >= 200 && statusCode < 300);
 }
 
+export const babelCompileCache = LRU<string>(<LRU.Options<string>>{
+  max: (10 * Math.pow(2, 20)),  // 10MB Cache Size
```
rictic: recommend bumping this up to 50 MB, or more

justinfagnani: Is 10MB enough? Maybe we want a flag for this.

src/compile-middleware.ts
```diff
@@ -66,6 +67,11 @@ function isSuccessful(response: Response) {
   return (statusCode >= 200 && statusCode < 300);
 }
 
+export const babelCompileCache = LRU<string>(<LRU.Options<string>>{
```
rictic: Why is the cast to `LRU.Options<string>` needed?

src/test/compile-middleware_test.ts
```diff
@@ -0,0 +1,128 @@
...
+      // Set the cached value to something else.
+      babelCompileCache.set(uncompiledHtml, 'IM IN UR CACHE');
+
+      // Make the request again, this time expect the value in cache as result.
+      const responseFromCache = await request();
+      assert.equal(responseFromCache.text, 'IM IN UR CACHE');
+
+      // Set the max size of the cache to something small and evict the cached
+      // value due to max cache length exceeded.
+      const originalCacheMax = babelCompileCache['max'];
```
justinfagnani: I might make this a separate test.

src/test/compile-middleware_test.ts
```diff
@@ -0,0 +1,128 @@
...
+      // Set the cached value to something else.
+      babelCompileCache.set(uncompiledJs, 'IM IN UR CACHE');
+
+      // Make the request again, this time expect the value in cache as result.
+      const responseFromCache = await request();
+      assert.equal(responseFromCache.text, 'IM IN UR CACHE');
+
+      // Set the max size of the cache to something small and evict the cached
+      // value due to max cache length exceeded.
+      const originalCacheMax = babelCompileCache['max'];
```
justinfagnani: ditto

src/test/compile-middleware_test.ts
```diff
@@ -0,0 +1,128 @@
...
+      assert.equal(babelCompileCache.has(uncompiledHtml), false);
+
+      // Request again and expect the output to match the original compiled
+      // response.
+      const responseFromCacheMiss = await request();
+      assert.equal(responseFromCacheMiss.text, responseAfterCompile.text);
+    });
+
+    test('caches javascript compilation results', async() => {
+      const app =
```
justinfagnani: can you make one test function that you pass different path / contents to?

justinfagnani:
Nice, simple PR :)

src/compile-middleware.ts
usergenic: You know, it might not be.  I needed to do it *before* I set the type on `LRU<string>` -- maybe that removes the need...  meh.  I get this if I don't have `<LRU.Options<string>>`
```
Argument of type '{ max: number; length: (n: string, key: string) => number; }' is not assignable to parameter of type 'number'
```

Maybe @types/lru-cache.d.ts declaration is bad.  And it should feel bad.

src/compile-middleware.ts
usergenic: Done.

src/test/compile-middleware_test.ts
usergenic: Yeah I debated that with myself, but lost because lazy.  Will separate.

src/compile-middleware.ts
usergenic: A cli option would be fairly easy to add.  I'll accept the size in bytes so I don't have to translate it.

src/compile-middleware.ts
```diff
@@ -66,6 +67,13 @@ function isSuccessful(response: Response) {
   return (statusCode >= 200 && statusCode < 300);
 }
 
+// NOTE: To change the max length of the cache at runtime, just use bracket
+// notation, i.e. `babelCompileCache['max'] = 64 * 1024` for 64KB limit.
+export const babelCompileCache = LRU<string>(<LRU.Options<string>>{
```
FredKSchott: this notation is a little crazy, why is the `<LRU.Options<string>>` cast needed? Is that needed to denote that you are configuring a cache of strings?

src/compile-middleware.ts
```diff
@@ -83,7 +91,10 @@ export function babelCompile(forceCompile: boolean): RequestHandler {
       const compile = forceCompile || needCompilation(uaParser);
 
       if (compile) {
```
FredKSchott: Lots of unnecessary branching going on here, can you change this to a `if(!compile) { return body; }` to exit early and reduce nesting?

src/compile-middleware.ts
```diff
@@ -100,10 +111,14 @@ export function babelCompile(forceCompile: boolean): RequestHandler {
               console.warn(`Error compiling script in ${request.path}: ${e}`);
             }
           }
-          return parse5.serialize(document);
+          const compiledHtml = parse5.serialize(document);
+          babelCompileCache.set(body, compiledHtml);
+          return compiledHtml;
         } else if (javaScriptMimeTypes.indexOf(contentType) >= 0) {
```
FredKSchott: likewise, else is unnecessary here with return statement above

src/test/compile-middleware_test.ts
usergenic: Separated.

src/test/compile-middleware_test.ts
usergenic: Tests are separated and small enough now without adding an abstraction.

src/compile-middleware.ts
usergenic: I think its a side-effect of the typings file having two declarations for babelCompileCache (one takes a `number`, the other takes an `LRU.Options<T>`) and for some reason typescript can't figure out that my object matches the `LRU.Options` type without explicitly typing it out, even when I say `LRU<string>` it says my `{...}` isn't a number and makes me feel sad.

src/compile-middleware.ts
usergenic: I made it better.

usergenic:
@FredKSchott @justinfagnani Feedback addressed PTAL

src/compile-middleware.ts
usergenic: Added cli option.  Also moved default value to there.

src/compile-middleware.ts
```diff
@@ -83,36 +91,44 @@ export function babelCompile(forceCompile: boolean): RequestHandler {
...
+    try {
+      const script = dom5.getTextContent(scriptTag);
+      const compiledScriptResult = compileScript(script);
+      dom5.setTextContent(scriptTag, compiledScriptResult);
+    } catch (e) {
+      // By not setting textContent we keep the original script, which
+      // might work. We may want to fail the request so a better error
+      // shows up in the network panel of dev tools. If this is the main
+      // page we could also render a message in the browser.
+      console.warn(`Error compiling script in ${location}: ${e}`);
```
FredKSchott: If we're catching a parse/compile error thrown by babel, log `e.message` if it exists (stack trace would be unnecessary to the user, message would be easier to read)

src/compile-middleware.ts
```diff
@@ -83,36 +91,44 @@ export function babelCompile(forceCompile: boolean): RequestHandler {
...
-        } else if (javaScriptMimeTypes.indexOf(contentType) >= 0) {
-          try {
-            return compileScript(body);
-          } catch (e) {
-            console.warn(`Error compiling script in ${request.path}: ${e}`);
-          }
+          body = compileHtml(source, request.path);
+        }
+        if (javaScriptMimeTypes.indexOf(contentType) !== -1) {
+          body = compileScript(source);
```
FredKSchott: Is there a reason we catch compile errors below but not here? Consider wrapping in try-catch

src/compile-middleware.ts
usergenic: Changed!  Good idea.

usergenic:
@FredKSchott Can you approve PR?  If you think we should catch compile errors, lets do as separate PR after identifying what polyserve should do.  I want this PR to be behaviorally invisible/distinct except for adding caching.

PR #148, opened 2016-12-17
Move compilation from make_app to start_server to compile apps and we…

justinfagnani:

…ll as components


src/test/start_server_test.ts
```diff
@@ -75,6 +79,80 @@ suite('startServer', () => {
...
+          result: 'uncompiled',
+        }));
+
+    test(
+        'compiles external JS when --compile=auto and agent is unknown',
+        testCompilation({
+          url: '/components/test-component/test.js',
+          compile: 'auto',
+          result: 'compiled',
+        }));
```
rictic: Looks like there are no tests of compilation outside of `/components/`

justinfagnani: done

PR #163, opened 2017-03-20
Added tests to demonstrate User-Agent code wrt needCompilation behavior.

usergenic:
- I wanted to verify that we are handling of user-agent by the compile middleware, so I just added a couple of tests.

### DIFF
--- a/src/test/compile-middleware_test.ts
+++ b/src/test/compile-middleware_test.ts
@@ -108,5 +108,60 @@ suite('compile-middleware', () => {
       // The valid script tag should still be compiled.
       assert.notInclude(response.text, `<script>\nclass A {}\n</script>`);
     });
+
+    suite('with compile option set to \'auto\'', () => {
+
+      beforeEach(() => {
+        app = getApp({
+          root: root,
+          compile: 'auto',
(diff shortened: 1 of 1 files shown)


usergenic:
This is an old PR I never requested review of.  Just added tests for the auto compile behavior.

src/test/compile-middleware_test.ts
```diff
@@ -108,5 +108,60 @@ suite('compile-middleware', () => {
...
+          root: root,
+          compile: 'auto',
+          componentDir: path.join(root, 'bower_components'),
+        });
+        // Ensure a fresh cache for each test.
+        babelCompileCache.reset();
+      });
+
+      const userAgentsThatSupportES2015 = [
+        'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.98 Safari/537.36',
```
justinfagnani: @aomarks added some UA strings and whether they should trigger compilation in browserNeedsCompilation. Maybe reuse those now?

usergenic:
Unified the collections of user agents that need/don't-need compilation in the test file.

PR #174, opened 2017-04-12
Check Edge versions correctly in compilation switch.

### LINKED ISSUE #161: Always compile for MSEdge <= 38
In Microsoft Edge 38.14393.0.0 (Microsoft EdgeHTML 14.14393), ES6 constructor returns class instead of an instance (see https://github.com/Microsoft/ChakraCore/issues/1496).
In this case polyserve should compile resources.

aomarks:
Really fixes #161. This time for real! My previous PR #173 assumed that
the Edge user agent used the main browser release version. Actually it
uses the EdgeHTML version. The fix we care about for the original issue
is in a major release of Edge, but a minor release of EdgeHTML, so we
have to check minor version now too.

https://en.wikipedia.org/wiki/Microsoft_Edge#Release_history

Also added a test and a small refactor so that we can paste real user
agent strings.

 - [x] CHANGELOG.md already updated in #173

### DIFF
--- a/src/compile-middleware.ts
+++ b/src/compile-middleware.ts
@@ -87,8 +87,8 @@ export function babelCompile(forceCompile: boolean): RequestHandler {
 
     transform(request: Request, response: Response, body: string): string /**/ {
       const contentType = getContentType(response);
-      const uaParser = new UAParser(request.headers['user-agent']);
-      const compile = forceCompile || needCompilation(uaParser);
+      const compile =
+          forceCompile || needCompilation(request.headers['user-agent']);
 
       if (compile) {
         const source = body;
@@ -141,15 +141,19 @@ const isInlineJavaScript = dom5.predicates.AND(
     dom5.predicates.hasTagName('script'),
     dom5.predicates.NOT(dom5.predicates.hasAttr('src')));
 
-function needCompilation(uaParser: UAParser): boolean {
-  const browser = uaParser.getBrowser();
+export function needCompilation(userAgent: string): boolean {
+  const browser = new UAParser(userAgent).getBrowser();
   const versionSplit = browser.version && browser.version.split('.');
   const majorVersion = versionSplit ? parseInt(versionSplit[0], 10) : -1;
+  const minorVersion = versionSplit && versionSplit.length > 1 ?
+      parseInt(versionSplit[1], 10) :
+      -1;
 
   const supportsES2015 = (browser.name === 'Chrome' && majorVersion >= 49) ||
       (browser.name === 'Chromium' && majorVersion >= 49) ||
       (browser.name === 'Safari' && majorVersion >= 10) ||
-      (browser.name === 'Edge' && majorVersion >= 40) ||
+      (browser.name === 'Edge' &&
+       (majorVersion > 15 || (majorVersion === 15 && minorVersion >= 15063))) ||
       (browser.name === 'Firefox' && majorVersion >= 51);
   return !supportsES2015;
 }
--- a/src/compile-middleware.ts
+++ b/src/compile-middleware.ts
@@ -149,7 +149,7 @@ function needCompilation(uaParser: UAParser): boolean {
   const supportsES2015 = (browser.name === 'Chrome' && majorVersion >= 49) ||
       (browser.name === 'Chromium' && majorVersion >= 49) ||
       (browser.name === 'Safari' && majorVersion >= 10) ||
-      (browser.name === 'Edge' && majorVersion >= 14) ||
+      (browser.name === 'Edge' && majorVersion >= 40) ||
       (browser.name === 'Firefox' && majorVersion >= 51);
   return !supportsES2015;
 }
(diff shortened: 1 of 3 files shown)


src/compile-middleware.ts
```diff
@@ -141,15 +141,19 @@ const isInlineJavaScript = dom5.predicates.AND(
     dom5.predicates.hasTagName('script'),
     dom5.predicates.NOT(dom5.predicates.hasAttr('src')));
 
-function needCompilation(uaParser: UAParser): boolean {
-  const browser = uaParser.getBrowser();
+export function needCompilation(userAgent: string): boolean {
+  const browser = new UAParser(userAgent).getBrowser();
   const versionSplit = browser.version && browser.version.split('.');
   const majorVersion = versionSplit ? parseInt(versionSplit[0], 10) : -1;
+  const minorVersion = versionSplit && versionSplit.length > 1 ?
```
justinfagnani: **ES6 golf** this could be:
```javascript
const [majorVersion, minorVersion] = versionSplit.map((v) => parseInt(v, 10));
```

But we need to drop node v4 support, which we've started in other repos

src/compile-middleware.ts
```diff
@@ -141,15 +141,19 @@ const isInlineJavaScript = dom5.predicates.AND(
...
+      parseInt(versionSplit[1], 10) :
+      -1;
 
   const supportsES2015 = (browser.name === 'Chrome' && majorVersion >= 49) ||
       (browser.name === 'Chromium' && majorVersion >= 49) ||
       (browser.name === 'Safari' && majorVersion >= 10) ||
-      (browser.name === 'Edge' && majorVersion >= 40) ||
+      (browser.name === 'Edge' &&
+       (majorVersion > 15 || (majorVersion === 15 && minorVersion >= 15063))) ||
```
justinfagnani: I wonder if a link to https://en.wikipedia.org/wiki/Microsoft_Edge#Release_history would be helpful here

src/test/compile-middleware_test.ts
```diff
@@ -109,4 +109,32 @@ suite('compile-middleware', () => {
       assert.notInclude(response.text, `<script>\nclass A {}\n</script>`);
     });
   });
+
+  test('needCompilation', () => {
```
justinfagnani: 🥇

src/test/compile-middleware_test.ts
```diff
@@ -109,4 +109,32 @@ suite('compile-middleware', () => {
...
+        expected: false,
+      },
+      {
+        userAgent:
+            'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36 Edge/16.00000',
+        expected: false,
+      },
+    ];
+    for (const c of cases) {
+      assert.equal(needCompilation(c.userAgent), c.expected, c.userAgent);
```
justinfagnani: I find one downside of assertions/tests in plain loops to be that the line the error occurred on doesn't clearly indicate the failure. In this case the UA string is the only state needed, so it'd be easy to look up and find the test data, but also in this case I'm not sure the object representation above is more compact than an inline assertion.

No need to change though.

Also, just to be self-contradictory, another alternative I like is custom assertions (though I generally avoid Chai's recommended global modifying approach of `Assertion.addMethod()`):

```typescript
function assertNeedsCompilation(ua: string, expectation: boolean) {
  assert.equal(needCompilation(ua), expectation, ua);
}
assertNeedsCompilation('Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36 Edge/14.99999', true);
```

Some of the same downsides as a loop, unless there's a way to black-box the assertion function I don't know about.

aomarks: Yeah, I agree this table-driven testing pattern could lead to bad failure messages, but I do personally find it kind of straightforward to read. Maybe I'm just really used to it because I've been writing so much Go, though, where it's really common.

Latest commit has a slightly more concise version now that we can use destructuring. Happy to change to the function style if you prefer it, though.

src/compile-middleware.ts

src/compile-middleware.ts

justinfagnani:
Still LGTM. I'd rebase when merging.

valdrinkoshi:
Can we have this released & used in `polymer-cli`? What would be the ETA?
Needed for making tests 💚  for https://github.com/PolymerElements/iron-overlay-behavior/pull/218

aomarks:
0.17.0 is out now which has the edge fix.

aomarks:
A polymer-cli with the fix is out if you grab @next.

valdrinkoshi:
Tests are [still failing](https://travis-ci.org/PolymerElements/iron-overlay-behavior/builds/221192415), and that's because of `web-component-tester` which is still using the previous version of polyserve https://github.com/Polymer/web-component-tester/blob/master/package.json#L86 😭

valdrinkoshi:
Tracking issue: https://github.com/Polymer/web-component-tester/issues/528

PR #179, opened 2017-04-14
Inject Custom Elements ES5 Adapter into entry point under ES5.

### LINKED ISSUE #164: Inject es5 loader stanza into entrypoint html when serving es5 compiled JS.
See https://github.com/Polymer/polymer-cli/issues/627 for the proposed solution.

May need to do more than inject (i.e. may need to shuffle order-dependent imperatives like scripts, styles and html imports) all down into the body after the loader.

### LINKED ISSUE #170: JS is compiled for Chrome in device emulation mode
Chrome changes the UA to match the device it's emulating, and the in auto-compile mode we do the wrong thing...

aomarks:
- Fixes #164. Fixes #170.
- Bug fix for multiple transformer middlewares.
- Assume all transformer middlewares only operate on 200 responses.
- A little refactoring of the compiler middleware.

- [ ] Depends on https://github.com/Polymer/polymer-build/pull/181. Tests will not pass until that PR is merged and released as 1.1.0.


src/compile-middleware.ts
```diff
@@ -75,35 +71,33 @@ export const babelCompileCache = LRU<string>(<LRU.Options<string>>{
...
   return transformResponse({
-    shouldTransform(_request: Request, response: Response) {
-      return isSuccessful(response) &&
-          compileMimeTypes.indexOf(getContentType(response)) >= 0;
+    shouldTransform(request: Request, response: Response) {
+      // We must never compile the Custom Elements ES5 Adapter or other
+      // polyfills/shims.
+      const isPolyfill = url.parse(request.url)
+                             .pathname.split('/')
+                             .indexOf('webcomponentsjs') >= 0;
```
justinfagnani: You can use `Array#includes` instead of `indexof`. Also, rather than split the path then check the results, you could just do `pathname.includes('webcomponentsjs/')`

src/compile-middleware.ts
```diff
@@ -75,35 +71,33 @@ export const babelCompileCache = LRU<string>(<LRU.Options<string>>{
...
-        babelCompileCache.set(source, body);
+      const source = body;
+      const cached = babelCompileCache.get(source);
+      if (cached !== undefined) {
+        return cached;
+      }
+      if (contentType === htmlMimeType) {
+        body = compileHtml(source, request.path);
+      }
+      if (javaScriptMimeTypes.indexOf(contentType) !== -1) {
```
justinfagnani: change `indexOf` to `includes`

src/custom-elements-es5-adapter-middleware.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+import {parse as parseContentType} from 'content-type';
+import {Request, RequestHandler, Response} from 'express';
+import {addCustomElementsEs5Adapter} from 'polymer-build';
+
+import {browserNeedsCompilation} from './compile-middleware';
+import {transformResponse} from './transform-middleware';
+
+// Returns an express middleware that injects the Custom Elements ES5 Adapter
```
justinfagnani: SHould this be a `/** */` comment?

src/transform-middleware.ts
```diff
@@ -25,7 +25,9 @@ export function transformResponse(transformer: ResponseTransformer):
 
     function shouldTransform() {
```
justinfagnani: For clarity, since even I forgot I did this, can you mention that this function memoizes it's result

src/compile-middleware.ts
aomarks: Here's what the CLI does: https://github.com/Polymer/polymer-cli/blob/master/src/build/optimize-streams.ts#L73

Same idea here, although the CLI is looking at a file path (hence the check for two types of delimiters) while here we're looking at a URL.

Could we instead check specifically that the file name is `custom-elements-es5-adapter.js`? I wasn't sure if we need to be careful about not compiling anything from the entire package (which is what CLI seems to be assuming), or if it's really only this specific file that matters.

justinfagnani: We should avoid compiling everything from that package, it's all already very compiled with Closure. This is fine for now.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+* Inject the Custom Elements ES5 Adapter into the entry point when serving ES5.
```
aomarks: Done.

src/compile-middleware.ts
aomarks: I guess that would match "/notwebcomponentsjs/" though :D

src/compile-middleware.ts

src/custom-elements-es5-adapter-middleware.ts
aomarks: Done.

src/transform-middleware.ts
aomarks: Done.

src/compile-middleware.ts
aomarks: Ack.

PR #182, opened 2017-04-18
Ability for polyserve to serve entry points other than index.html

### LINKED ISSUE #180: polyserve should serve the entry point configured in polymer.json
Currently it assumes index.html.

aomarks:
- Part of #180.
- Note that it won't be possible to use the feature unless you are calling via polymer-cli.

### DIFF
--- a/src/start_server.ts
+++ b/src/start_server.ts
@@ -16,7 +16,7 @@ import * as assert from 'assert';
 import * as express from 'express';
 import * as fs from 'mz/fs';
 import * as path from 'path';
-import {urlFromPath} from 'polymer-build';
+import {urlFromPath} from 'polymer-build/lib/path-transformers';
 import * as send from 'send';
 // TODO: Switch to node-http2 when compatible with express
 // https://github.com/molnarg/node-http2/issues/100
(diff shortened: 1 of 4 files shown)


justinfagnani:
If this requires a change to polymer-build, doesn't it also need to update the polymer-build version constraint? Can you add a link to the PR or issues this depends on?

LGTM other than that.

src/start_server.ts
```diff
@@ -16,6 +16,7 @@ import * as assert from 'assert';
 import * as express from 'express';
 import * as fs from 'mz/fs';
 import * as path from 'path';
+import {urlFromPath} from 'polymer-build';
```
justinfagnani: It is a little misplaced, but I don't really have a better idea at the moment. We don't have a library of utilities like this. Maybe just don't export out of the main polymer-build module, but import directly from `polymer-build/path-transformers`

aomarks: SGTM Thanks!

aomarks:
> If this requires a change to polymer-build, doesn't it also need to update the polymer-build version constraint? Can you add a link to the PR or issues this depends on?
> 
> LGTM other than that.

As discussed, importing directly from path-transformers now, so no dependency changes needed after all.

PR #198, opened 2017-05-23
Don't exclude webcomponentsjs polyfill *tests* from compilation.

aomarks:
Includes a clang format commit, 2nd commit is the interesting one.

Also prepare release of 0.19.1

- [x CHANGELOG.md has been updated


aomarks:
Master was already broken for tsc reasons. Fixed that too.

PR #203, opened 2017-06-23
Allow for nested packages (package names with a /)

justinfagnani:
Fixes https://github.com/Polymer/polyserve/issues/204


.travis.yml
```diff
@@ -2,5 +2,6 @@ language: node_js
 node_js:
   - "node"
   - "6"
+before_install: if [[ `npm -v` != 5* ]]; then npm i -g npm@5; fi
```
aomarks: this will be weird when npm 6 comes out. can you just always install npm latest instead?

src/make_app.ts
```diff
@@ -102,3 +103,12 @@ export function makeApp(options: AppOptions): PolyserveApplication {
   app.packageName = packageName;
   return app;
 }
+
+function arrayStartsWith(path: any[], prefix: any[]) {
```
aomarks: s/path/arr ?

.travis.yml
justinfagnani: true, and yes

src/make_app.ts
justinfagnani: done

PR #205, opened 2017-06-26
Add —npm option

justinfagnani:
This adds an `--npm` flag that sets the component folder to `node_modules` (`-c` flag), and reads the package name from package.json rather than bower.json.

There's also a little refactoring of the config reading.


test/combo-package/bower.json
```diff
@@ -0,0 +1,3 @@
+{
+  "name": "a-bower-package"
+}
```
aomarks: you should tell your editor to add trailing newlines

justinfagnani: done :)

PR #209, opened 2017-08-03
Fix node type errors

FredKSchott:
Node type definitions changed under our feet a bit since last change. Added fixes and some comments to explain.

### DIFF
--- a/src/compile-middleware.ts
+++ b/src/compile-middleware.ts
@@ -62,7 +62,7 @@ const compileMimeTypes = [
 export const isPolyfill = /(^|\/)webcomponentsjs\/[^\/]+$/;
 
 function getContentType(response: Response) {
-  const contentTypeHeader = response.getHeader('Content-Type');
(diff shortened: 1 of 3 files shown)


src/compile-middleware.ts
```diff
@@ -62,7 +62,7 @@ const compileMimeTypes = [
 export const isPolyfill = /(^|\/)webcomponentsjs\/[^\/]+$/;
 
 function getContentType(response: Response) {
-  const contentTypeHeader = response.getHeader('Content-Type');
+  const contentTypeHeader = response.getHeader('Content-Type') as string;
```
rictic: what changed about the header types?

usergenic: I fixed these a different way in https://github.com/Polymer/polyserve/pull/210

I converted the getHeader('x') calls to get('x') which had the correct typings info and this wasn't necessary.

PR #210, opened 2017-08-07
Upgrade polymer build to 2

usergenic:
- Update Polymer Build to v2.0.0

Had some typings related issues I was able to fix by using aliased public methods .getHeader -> .get on Request+Response objects.


usergenic:
Note this update to polyserve will enable polymer-cli to ship only one version of polymer-build and bundler instead of multiple versions.

PR #213, opened 2017-08-15
Transform ES modules to AMD/RequireJS

### LINKED ISSUE #208: Add module transform to compilation
We need to support JS modules. This includes:

- Compiling JS modules with one of the Babel module transforms.
- Supporting a loader for that transform, possibly injecting it into entrypoints.
- Transform HTML files to remove `type=module` (important in cases where we send compiled code to browsers that support modules).

AMD is probably the easiest module format to target. If all import specifiers are paths (because they are assumed to work with native modules), then the AMD registry/loader can be very simple. Possibly based on [IMD](https://github.com/PolymerLabs/imd), or we could just use RequireJS.

aomarks:
Fixes #208
- [x] CHANGELOG updated.

### DIFF
--- a/src/compile-middleware.ts
+++ b/src/compile-middleware.ts
@@ -22,7 +22,17 @@ import * as parse5 from 'parse5';
 
 import {transformResponse} from './transform-middleware';
 
-const babelTransformers = [
+const p = dom5.predicates;
+
+const isJsScriptNode = p.AND(
+    p.hasTagName('script'),
+    p.OR(
+        p.NOT(p.hasAttr('type')),
+        p.hasAttrValue('type', 'text/javascript'),
+        p.hasAttrValue('type', 'application/javascript'),
+        p.hasAttrValue('type', 'module')));
+
+const es2015Plugins = [
   'babel-plugin-transform-es2015-arrow-functions',
   'babel-plugin-transform-es2015-block-scoped-functions',
   'babel-plugin-transform-es2015-block-scoping',
@@ -44,6 +54,10 @@ const babelTransformers = [
   'babel-plugin-transform-regenerator',
 ].map((name) => require(name));
 
+const modulesPlugins = [
+  'babel-plugin-transform-es2015-modules-amd',
+].map((name) => require(name));
+
 const javaScriptMimeTypes = [
   'application/javascript',
   'application/ecmascript',
@@ -55,7 +69,13 @@ const htmlMimeType = 'text/html';
 
 const compileMimeTypes = [
   htmlMimeType,
-].concat(javaScriptMimeTypes);
+  ...javaScriptMimeTypes,
+];
+
+interface compileOptions {
+  transformES2015: boolean;
+  transformModules: boolean;
+}
 
 // Match the polyfills from https://github.com/webcomponents/webcomponentsjs,
 // but not their tests.
@@ -72,7 +92,6 @@ export const babelCompileCache = LRU<string>(<LRU.Options<string>>{
   length: (n: string, key: string) => n.length + key.length
 });
 
-
 export function babelCompile(forceCompile: boolean): RequestHandler {
   return transformResponse({
     shouldTransform(request: Request, response: Response) {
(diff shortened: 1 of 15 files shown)


src/compile-middleware.ts
```diff
@@ -84,56 +104,135 @@ export function babelCompile(forceCompile: boolean): RequestHandler {
...
+      if (transformingModule) {
+        plugins.push(...modulesPlugins);
+      }
+
+      const js = dom5.getTextContent(scriptTag);
+      let compiled;
+      try {
+        compiled = babelCore.transform(js, {plugins}).code;
+      } catch (e) {
+        // By not setting textContent we keep the original script, which might
```
justinfagnani: It might be nice to inject a "console.warn" into the script so there's an error in the browser's console

aomarks: Cleaned up the comment a bit and added a TODO. (Note this behavior was the same before this PR).

aomarks:
Added new require monkey patching and tests. PTAL.

src/compile-middleware.ts
```diff
@@ -84,56 +104,195 @@ export function babelCompile(forceCompile: boolean): RequestHandler {
...
+      }
+
+      dom5.setTextContent(scriptTag, compiled);
     }
   }
+
+  if (wctScriptTag && requireScriptTag) {
+    // This looks like a Web Component Tester script, and we have converted
+    // ES modules to AMD. Converting a module to AMD means that it will not
+    // execute synchronously like an ES module would. Since WCT listens for the
```
justinfagnani: "synchronously" isn't quite the right word here. DOMCOntentLoaded waits for modules to run is all.

src/compile-middleware.ts
```diff
@@ -84,56 +104,195 @@ export function babelCompile(forceCompile: boolean): RequestHandler {
...
+    // To address this, we inject a block of JS that uses WCT's `waitFor` hook
+    // to defer running tests until our AMD modules have loaded. Do this as
+    // late as possible, before the WCT script, because users may be setting
+    // their own `waitFor` that musn't clobber ours. Likewise we must call
+    // theirs if we find it.
+    dom5.insertBefore(
+        wctScriptTag.parentNode, wctScriptTag, parse5.parseFragment(`
+<script>
+  (function() {
+    window.WCT = window.WCT || {};
```
justinfagnani: I think the snippet itself should have a comment. At least that it was injected by polyserve.

src/compile-middleware.ts
```diff
@@ -84,56 +104,195 @@ export function babelCompile(forceCompile: boolean): RequestHandler {
...
+    var originalWaitFor = window.WCT.waitFor;
+    window.WCT.waitFor = function(cb) {
+      window._wctCallback = function() {
+        if (originalWaitFor) {
+          originalWaitFor(cb);
```
justinfagnani: Maybe explain this a bit because it's a little hard to reason about at first

src/compile-middleware.ts
```diff
@@ -84,56 +104,195 @@ export function babelCompile(forceCompile: boolean): RequestHandler {
...
+        } else {
+          cb();
+        }
+      };
+    };
+  }());
+</script>
+`));
+
+    // Monkey patch `require` to keep track of loaded AMD modules. Note this
```
justinfagnani: nice solution!

src/compile-middleware.ts
```diff
@@ -84,56 +104,195 @@ export function babelCompile(forceCompile: boolean): RequestHandler {
...
+    // that's an assumption WCT normally makes anyway. Do this right after
+    // RequireJS is loaded, and hence before the first module is registered.
+    dom5.insertAfter(
+        requireScriptTag.parentNode, requireScriptTag, parse5.parseFragment(`
+<script>
+  (function() {
+    var originalRequire = window.require;
+    var moduleCount = 0;
+    window.require = function(deps, factory) {
+      moduleCount++;
```
justinfagnani: I wonder if we should have a setTimeout here that checks if this require call's factory completed, and warns if not. That way there's at least something to go on if WCT doesn't start.

src/test/compile-middleware_test.ts
```diff
@@ -25,33 +25,31 @@ chai.use(chaiAsPromised);
...
   'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36 Edge/14.99999',
   'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36 Edge/15.14986',
 ];
 
-const userAgentsThatSupportES2015 = [
-  'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.98 Safari/537.36',
-  'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36 Edge/15.15063',
-  'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36 Edge/16.00000',
+const userAgentsThatSupportES2015AndModules = [
+  'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.39 Safari/537.36',
```
justinfagnani: Chrome 61 for now. We maybe have to bump this to 62

src/compile-middleware.ts

src/compile-middleware.ts

src/compile-middleware.ts

src/compile-middleware.ts
aomarks: Discussed offline, added a comment.

aomarks:
PTAL. RequireJS is now a dependency of Polyserve, and we'll serve it from Polyserve's node_modules if it's not found in the project's component dir.

src/compile-middleware.ts
```diff
@@ -84,56 +105,207 @@ export function babelCompile(forceCompile: boolean): RequestHandler {
...
+      }
+      if (transformingModule) {
+        plugins.push(...modulesPlugins);
+      }
+
+      let compiled;
+      try {
+        compiled = babelCore.transform(js, {plugins}).code;
+      } catch (e) {
+        // Leave the original script as-is. Maybe it will work...
```
usergenic: Unless the `let compiled` statement above is initialized with the original js value, it will be undefined and this would break on the `.replace` call below or the dom5.setTextContent would receive undefined.

aomarks: The replace call won't (and shouldn't) run because of the `continue`.

aomarks: Updated comment to bring attention to the continue.

PR #220, opened 2017-09-14
Resolve requirejs actual install path

usergenic:
- Fix issue where requirejs is installed somewhere other than in polyserve's own node_modules subfolder.

### DIFF
--- a/src/make_app.ts
+++ b/src/make_app.ts
@@ -34,9 +34,16 @@ export interface PolyserveApplication extends express.Express {
 // compile-middleware.ts), we inject a dependency on the RequireJS loader. We
 // include RequireJS as an NPM dependency of Polyserve so that users don't need
 // to make it a dependency of their project. This is our local path to the
-// library.
-const localRequirePath =
-    path.join(__dirname, '..', 'node_modules', 'requirejs', 'require.js');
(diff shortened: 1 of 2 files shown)


src/make_app.ts
```diff
@@ -34,9 +34,16 @@ export interface PolyserveApplication extends express.Express {
...
-    path.join(__dirname, '..', 'node_modules', 'requirejs', 'require.js');
+// library.  The resolved path of the `requirejs` package is actually its
+// `bin/r.js` file, so we have to truncate and change the path to get to the
+// `require.js` file we actually want.  Using a regexp here to guard against
+// a future version of RequireJS moving its main to a location at different
+// depth.
+// Resolved path: "somethingsomething/requirejs/bin/r.js"
+// Regexp match segment: "somethingsomething/requirejs/"
+const localRequirePath = path.join(
+    require.resolve('requirejs').match(/.*[/\\]requirejs[/\\]/)[0],
```
justinfagnani: can't this be replaced by `require.resolve('requirejs/require.js')`?

usergenic: O_o

OH BANANAS I think you are right.  Will fix.

For some reason I tried to `np` in polyserve to release new version and it published without its `lib` dir.  Don't know why.  Rushing to fix that first.

PR #227, opened 2017-09-26
Do not compile modules when nomodule script exists

### LINKED ISSUE #226: Do not transform modules when [nomodule] exists
Polyserve transfers modules into non-module scripts. However, if a user explicitly added a `<script nomodule ...>`, this results in running code twice if a browser does not support modules. A heuristic that would solve this would be to not transform the modules if a `[nomodule]` exists in the DOM AST.

TimvdLippe:
If a `<script nomodule>` exists in the file, do not transform modules. This prevents double instantiation of objects when both the compiled script and the non-compiled script are run.

Fixes #226

### DIFF
--- a/src/compile-middleware.ts
+++ b/src/compile-middleware.ts
@@ -145,6 +145,9 @@ function compileHtml(
 
   const jsNodes = dom5.queryAll(document, isJsScriptNode);
 
+  // Assume that if this document has a nomodule script, the author is already
+  // handling browsers that don't support modules, and we don't need to
+  // transform anything.
   if (jsNodes.find((node) => dom5.hasAttribute(node, 'nomodule'))) {
     return source;
   }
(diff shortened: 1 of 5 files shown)


src/compile-middleware.ts
```diff
@@ -143,7 +143,13 @@ function compileHtml(
   const document = parse5.parse(source);
   let requireScriptTag, wctScriptTag;
 
-  for (const scriptTag of dom5.queryAll(document, isJsScriptNode)) {
+  const jsNodes = dom5.queryAll(document, isJsScriptNode);
+
+  if (jsNodes.find((node) => dom5.hasAttribute(node, 'nomodule'))) {
```
aomarks: Add a brief comment? E.g. "Assume that if this document has a nomodule script, the author is already handling browsers that don't support modules, and we don't need to transform anything."

TimvdLippe: done

PR #233, opened 2017-10-26
Allow for relative roots that are resolved to the cwd

justinfagnani:

This allows for configurations like:

```ts
polyserve.makeApp({
  packageName: 'foo-bar',
  root: '../client',
})
```


justinfagnani:
Looks like I get different format-on-save than `npm run format`. Pushing fixed files

justinfagnani:
Thanks!

PR #245, opened 2018-01-15
Fixes XSS issue when rendering errors.

wesalvaro:
The request string used to detect this flaw was :
```
/<script>xss.nasl</script>.asp
```

The output was :

```
HTTP/1.1 404 Not Found
X-Powered-By: Express
Date: Tue, 09 Jan 2018 15:09:20 GMT
Connection: keep-alive
Transfer-Encoding: chunked


ENOENT: no such file or directory, stat '/path/<script>xss.nasl</script>.asp'
```

This change will set the content-type to an HTML document and escape the HTML contained.


wesalvaro:
@aomarks PTAL ;)

PR #246, opened 2018-02-05
Resolve and rewrite bare module specifiers.

justinfagnani:
Note: Still adding tests...


src/babel-plugin-bare-specifiers.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+import {dirname, relative} from 'path';
+import resolve = require('resolve');
+import {NodePath} from 'babel-traverse';
+import {ImportDeclaration, ExportNamedDeclaration, ExportAllDeclaration} from 'babel-types';
+
+const exportExtensions = require('babel-plugin-syntax-export-extensions');
+
+const isPathSpecifier = (s: string) => /^\.{0,2}\/.*/.test(s);
```
aomarks: the trailing `.*` could be removed, it's a noop

src/babel-plugin-bare-specifiers.ts
```diff
@@ -0,0 +1,68 @@
...
+
+type HasSpecifier =
+    ImportDeclaration|ExportNamedDeclaration|ExportAllDeclaration;
+
+/**
+ * Rewrites so-called "bare module specifiers" to be web-compatible paths.
+ */
+export const resolveBareSpecifiers =
```
aomarks: why not `export function`?

src/babel-plugin-bare-specifiers.ts
```diff
@@ -0,0 +1,68 @@
...
+    (filePath: string, isComponentRequest: boolean) => ({
```
aomarks: Is there a return type you can add to make it self-documenting? Not sure what this `{inherits, visitor, etc.}` interface is.

src/compile-middleware.ts
```diff
@@ -115,17 +115,28 @@ export function babelCompile(
         transformModules: forceCompile || !capabilities.has('modules'),
       };
       const optionsLeader = JSON.stringify(options);
-      const cached = babelCompileCache.get(optionsLeader + body);
+      const cached = babelCompileCache.get(optionsLeader + request.path + body);
```
aomarks: Can it be *just* `request.path` now? Add a comment?

src/compile-middleware.ts
```diff
@@ -115,17 +115,28 @@ export function babelCompile(
...
       if (cached !== undefined) {
         return cached;
       }
 
       let transformed;
       const contentType = getContentType(response);
+      let filePath = path.resolve(path.join(rootdir, request.path));
+      // The file path needs to include the filename for correct relative path
+      // calculation
+      const stat = fs.statSync(filePath);
```
aomarks: I guess we should avoid sync calls since we could be handling multiple requests? use fs-extra?

justinfagnani: `transform()` is sync, unfortunately. It really should be async.

src/compile-middleware.ts
justinfagnani: I'm really wondering why options got in there in the first place, since options should be unique for any running polyserve process. I tried to remove it but got a couple of weird failures. I added a TODO and filed https://github.com/Polymer/polyserve/issues/248 to track.

src/babel-plugin-bare-specifiers.ts
justinfagnani: `const` bindings are faster, if there's a default style it should be all exports are `const`.

src/babel-plugin-bare-specifiers.ts
justinfagnani: done

src/babel-plugin-bare-specifiers.ts
justinfagnani: I looked for a bit and couldn't find what package this type might be defined in.

PR #254, opened 2018-03-20
Fix import specifier rewriting when importing a dependency from a top-level module

justinfagnani:


src/compile-middleware.ts
```diff
@@ -116,11 +116,21 @@ export function babelCompile(
 
       let transformed;
       const contentType = getContentType(response);
-      const isComponentRequest = request.baseUrl === `/${componentUrl}` &&
-          !request.path.startsWith(`/${packageName}`);
+      let requestPath = request.path;
+      const isRootPathRequest = request.path.startsWith(`/${packageName}`);
```
aomarks: i guess we should make this startWith safer too?

justinfagnani: existing code? This logic shouldn't change on windows, so I'm going to defer for now.

aomarks: just add a todo?

justinfagnani: done

PR #258, opened 2018-03-21
Use the new componentRequestRelationship option, add tests

FredKSchott:
- [X] CHANGELOG.md has been updated
- Tests added / restructured a bit
- Needs latest polymer-build before it passes


src/compile-middleware.ts
```diff
@@ -123,6 +123,11 @@ export function babelCompile(
       // TODO(justinfagnani): Use path-is-inside, but test on Windows
       const isRootPathRequest = request.path.startsWith(`/${packageName}`);
       const isComponentRequest = request.baseUrl === `/${componentUrl}`;
+      let componentRequestRelationship: undefined|'../../'|'../';
+      if (isComponentRequest) {
+        const isScopedComponent = packageName.startsWith('@');
```
justinfagnani: related to the comment in polymer-build, if you just pass the package name, you shouldn't have to rely on special characters or an assumption that there's only one or two levels of scopes.

test/golden/npm-package/index.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html><html><head>
     <script type="module">
-import * as test from './node_modules/test-dependency/dep.js';</script>
+import * as test from '../test-dependency/dep.js';</script>
```
FredKSchott: @justinfagnani confirm these changes are okay? This looks like it was caused by https://github.com/Polymer/polymer-build/pull/320 which is getting pulled into the next release

justinfagnani: this file was being loaded as an app-style request, so the URL should really be `./node_modules/...`

